PR #2, opened 2023-04-03
[#IOPAE-268] Add Open API specification

giuseppedipinto:
Add Open API v3 specification

#### List of Changes
- add _openapi.yaml_ in root directory

#### Motivation and Context

Describe all the functionalities that will be exposed.

#### Checklist:

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -14,7 +14,7 @@ info:
   contact:
     name: PagoPA S.p.A.
     url: https://docs.pagopa.it/io-guida-tecnica/
-  version: 0.4.0
+  version: 0.5.0
 servers:
   - url: https://todo-url
 tags:
@@ -26,6 +26,8 @@ tags:
     description: Service Release API specification
   - name: services-authorization
     description: Services Authorization API specification
+  - name: service-history
+    description: Service History API specification
   - name: organizations
     description: Organizations API specification
 paths:
@@ -441,6 +443,39 @@ paths:
           description: Internal server error
       security:
         - api_key: []
+  /services/{serviceId}/history:
+    get:
+      tags:
+        - service-history
+      summary: Retrieve service history
+      description: Retrieve all versions of a service
+      operationId: getServiceHistory
+      parameters:
+        - name: serviceId
+          in: path
+          description: ID of the service
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Service history fetched successfully
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Service'
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error
+      security:
+        - api_key: []
   /organizations/{fiscalCode}/logo:
     put:
       tags:
@@ -575,6 +610,16 @@ components:
       required: 
         - scope
     ServiceStatus:
+      type: object
+      properties: 
+        value:
+            $ref: '#/components/schemas/ServiceStatusType'
+        reason:
+          description: Reason for status value. This property is optional and exists only when value is _rejected_
+          type: string
+      required: 
+        - value
+    ServiceStatusType:
       description: |
         Service status lifecycle<br>
         - _draft_: A new draft of the service has been created, filled in completely or partially and saved by the system.
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1,0 +1,712 @@
+openapi: 3.0.3
+info:
+  title: IO Services CMS
+  description: |-
+    _This is the IO Service CMS based on the OpenAPI 3.0 specification._
+
+    ## Introduction
+    This is the documentation of the IO API for 3rd party services. This API
+    enables Public Administration services to integrate with the IO platform. IO
+    enables services to communicate with Italian citizens via the [IO
+    app](https://io.italia.it/).
+
+    ### How to get Manage API Key
+    To get access to this API, you'll need to register on the [PagoPA SelfCare Platform](https://selfcare.pagopa.it/).
+
+    _TODO: registration process?_
+   
+    ### Allowed Markdown formatting
+    Not all Markdown formatting is currently available. Currently you can use
+    the following formatting:
+
+      * Headings
+      * Text stylings _(bold, italic, etc...)_
+      * Lists _(bullet and numbered)_
+    
+    ### Some useful links:
+    - [Github public codebase](https://github.com/pagopa/io-services-cms)
+    - [The source API definition for IO Services CMS](https://github.com/pagopa/io-services-cms/openapi/openapi.yaml)
+
+  termsOfService: http://url/terms/
+  contact:
+    name: PagoPA S.p.A.
+    url: https://docs.pagopa.it/io-guida-tecnica/
+  version: 0.1.0
+servers:
+  - url: https://todo-url
+tags:
+  - name: services
+    description: Services API specification
+  - name: service-actions
+    description: Service Actions API specification
+  - name: service-release
+    description: Service Release API specification
+  - name: services-authorization
+    description: Services Authorization API specification
+  - name: organizations
+    description: Organizations API specification
+paths:
+  /services:
+    post:
+      tags:
+        - services
+      summary: Create a new service
+      description: |
+        Create a new Service with the attributes provided in the request payload.<br><br>_Note: This API operation needs the __Manage API key__
+        owned by the service user, otherwise `403 forbidden` will be returned to the caller._
+      operationId: createService
+      requestBody:
+        description: |
+          A service body payload
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ServicePayload'
+        required: true
+      responses:
+        '200':
+          description: Successful operation _(service created)_
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Service'
+        '400':
+          description: Invalid payload
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service cannot be created)_
+      security:
+        - subscription_key: []
+    get:
+      tags:
+        - services
+      summary: Retrieve all services
+      description: Retrieve all services owned by the calling user
+      operationId: getServices
+      responses:
+        '200':
+          description: Successful operation _(service list)_
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Service'
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '429':
+          description: Too many requests
+      security:
+        - subscription_key: []
+  /services/{serviceId}:
+    get:
+      tags:
+        - services
+      summary: Retrieve service
+      description: Retrieve a service by `serviceId`
+      operationId: getService
+      parameters:
+        - name: serviceId
+          in: path
+          description: ID of the service
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful operation _(service data)_
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Service'
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+      security:
+        - subscription_key: []
+    put:
+      tags:
+        - services
+      summary: Update service
+      description: |
+        Update an existing service by by `serviceId`. <br><br>_Note: This API operation needs the __Manage API key__ owned by the service user, otherwise `403 forbidden` will be returned to the caller._
+      operationId: updateService
+      parameters:
+        - name: serviceId
+          in: path
+          description: ID of the service
+          required: true
+          schema:
+            type: string
+      requestBody:
+        description: Updated service payload
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ServicePayload'
+        required: true
+      responses:
+        '200':
+          description: Successful operation _(service updated)_
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Service'
+        '400':
+          description: Invalid payload
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service cannot be updated)_
+      security:
+        - subscription_key: []
+    delete:
+      tags:
+        - services
+      summary: Delete service
+      description: |
+        Delete a service by `serviceId`. <br>This operation is a logical deletion: the service persist in user services, but it's status become
+        `deleted`<br>_Note: This API operation needs the __Manage
+        API key__ owned by the service user, otherwise `403 forbidden` will be returned to the caller._
+      operationId: deleteService
+      parameters:
+        - name: serviceId
+          in: path
+          description: ID of the service
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful operation _(service deleted)_
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service cannot be deleted)_
+      security:
+        - subscription_key: []
+  /services/{serviceId}/logo:
+    put:
+      tags:
+        - services
+      summary: Upload service logo
+      description: |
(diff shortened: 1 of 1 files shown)


rodoviti:
@giuseppedipinto if I load [openapi.yaml](https://github.com/pagopa/io-services-cms/blob/IOPAE-268--OpenAPI-specification/openapi.yaml) into a [Swagger Editor](https://editor.swagger.io/) i get some errors...
image

giuseppedipinto:
@rodoviti errors on lines 705-706 are not highlighted in new version of swagger editor, however we can delete those empty objects.

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service key cannot be regenerated)_
+      security:
+        - subscription_key: []
+  /services/{serviceId}/actions/review:
+    patch:
```
rodoviti, line 295: I am not very "happy" with the design of "actions" because they are not RESTful API.

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '412':
+          description: Precondition failed _(service status is incompatible with review action request)_
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service review cannot be taken in charge)_
+      security:
+        - subscription_key: []
+  /services/{serviceId}/actions/explain:
+    patch:
```
rodoviti, line 332: I am not very "happy" with the design of "actions" because they are not RESTful API.

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+      parameters:
+        - name: serviceId
+          in: path
+          description: ID of the service
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful operation _(service unpublished)_
```
rodoviti, line 449: I propose the following response code:
```suggestion
        '204':
          description: Service unpublished
```

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(organization logo cannot be updated)_
```
rodoviti, line 498: Why "_organization logo cannot be updated_"?

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service cannot be unpublished)_
```
rodoviti, line 459: Why "_service cannot be unpublished_"?

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+      requestBody:
+        description: Service key type
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/SubscriptionKeyTypePayload'
+        required: true
+      responses:
+        '200':
+          description: Successful operation _(service key regenerated)_
```
rodoviti, line 275:
```suggestion
          description: Service key regenerated
```

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service key cannot be regenerated)_
```
rodoviti, line 291: Why "_service key cannot be regenerated_"?

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '412':
+          description: Precondition failed _(service status is incompatible with release request)_
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service cannot be published)_
```
rodoviti, line 397: Why "_service cannot be published_"?

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+                $ref: '#/components/schemas/Service'
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+      security:
```
rodoviti, line 137:
```suggestion
        '500':
          description: Internal server error
      security:
```

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service cannot be updated)_
```
rodoviti, line 178: Why "_service cannot be updated_"?

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+                $ref: '#/components/schemas/Service'
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+      security:
```
rodoviti, line 430:
```suggestion
        '500':
          description: Internal server error
      security:
```

openapi.yaml
```diff
@@ -0,0 +1,710 @@
...
+        '400':
+          description: Invalid payload
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service cannot be created)_
```
rodoviti, line 82: Why "_service cannot be created_"?

openapi.yaml
```diff
@@ -0,0 +1,710 @@
...
+        - subscription_key: []
+    get:
+      tags:
+        - services
+      summary: Retrieve all services
+      description: Retrieve all services owned by the calling user
+      operationId: getServices
+      responses:
+        '200':
+          description: Successful operation _(service list)_
```
rodoviti, line 93:
```suggestion
          description: Services fetched
```

openapi.yaml
```diff
@@ -0,0 +1,710 @@
...
+      parameters:
+        - name: serviceId
+          in: path
+          description: ID of the service
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful operation _(service data)_
```
rodoviti, line 124:
```suggestion
          description: Service fetched
```

openapi.yaml
```diff
@@ -0,0 +1,710 @@
...
+      requestBody:
+        description: Updated service payload
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ServicePayload'
+        required: true
+      responses:
+        '200':
+          description: Successful operation _(service updated)_
```
rodoviti, line 162:
```suggestion
          description: Service updated
```

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '500':
+          description: Internal server error _(service cannot be updated)_
+      security:
+        - subscription_key: []
+    delete:
+      tags:
+        - services
+      summary: Delete service
+      description: |
+        Delete a service by `serviceId`. <br>This operation is a logical deletion: the service persist in user services, but it's status become
```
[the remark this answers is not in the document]
balanza: I agree with @rodoviti, and I add a question (possibly not related with this PR): what if I `getService` on a deleted item? Would it be:
1. 404 Not Found
2. 200 { status: DELETED }

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service cannot be deleted)_
```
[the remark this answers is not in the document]
balanza: Right. Generally speaking we should have:
* Operations that cannot be completed because of data (provided or persisted) --> 4xx
* Operations that failed for technical issues --> 5xx

The meaning for the user is, respectively, "I'm doing something wrong, let's find out" and "The system is having trouble, let's try again in a while".

openapi.yaml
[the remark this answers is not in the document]
rodoviti: @balanza for me should return a 404 Not Found, but it might depend on the Use Cases.
I also add that this operation could (or should?!) be idempotent

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '404':
+          description: Not found
+        '429':
+          description: Too many requests
+        '500':
+          description: Internal server error _(service logo cannot be updated)_
```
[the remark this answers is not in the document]
giuseppedipinto: As on previous similar comment, we will have "Internal server error" meaning a techincal failure

openapi.yaml
giuseppedipinto: Same as previous similar comment

openapi.yaml
giuseppedipinto: Same as previous similar comment

openapi.yaml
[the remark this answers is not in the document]
giuseppedipinto: _404_ should be more user "transparent", but they loose the possibility of knowing if a service is deleted or not and in this case _200 { status: DELETED }_ is a good opportunity.
We have to choose...

openapi.yaml
```diff
@@ -0,0 +1,712 @@
...
+    description: Organizations API specification
+paths:
+  /services:
+    post:
+      tags:
+        - services
+      summary: Create a new service
+      description: |
+        Create a new Service with the attributes provided in the request payload.<br><br>_Note: This API operation needs the __Manage API key__
+        owned by the service user, otherwise `403 forbidden` will be returned to the caller._
```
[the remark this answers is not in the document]
giuseppedipinto: We have completely removed this description. We'll discuss it later.

openapi.yaml
giuseppedipinto: We think that Option 1. is what we currently have.
We agree that "services-actions" could be improved. We suggest:

__services-actions:__
__from__ PATCH /services/{serviceId}/actions/review __to__ POST /services/{serviceId}/review _(Submit a review)_
__from__ PATCH /services/{serviceId}/actions/explain __to__ PATCH/PUT /services/{serviceId}/review _(Explain a review)_

openapi.yaml
giuseppedipinto: see comment above

openapi.yaml
rodoviti: I add one more "trade-off" point:
4. we consider the service status a "sub-resources", so we can update it by:
  - `PUT /services/{serviceId}/status/{status}` where `status ∈ {Submitted, Rejected, ecc.}`

I mean, if this isn't the "external" but a kind of "internal" OpenAPI definition, I prefer to design it as point 4, because is more clear to the "domain team". On the other hand, for the "external/proxy" OpenAPI definition, i think the  2nd point is more "understanding"

openapi.yaml
```diff
@@ -0,0 +1,730 @@
...
+        '500':
+          description: Internal server error
+      security:
+        - api_key: []
+    put:
+      tags:
+        - services
+      summary: Update service
+      description: |
+        Update an existing service by by `serviceId`.
```
rodoviti, line 143:
```suggestion
      description: Update an existing service by its id.
```

openapi.yaml
```diff
@@ -0,0 +1,730 @@
...
+          description: Too many requests
+        '500':
+          description: Internal server error
+      security:
+        - api_key: []
+    get:
+      tags:
+        - services
+      summary: Retrieve all services
+      description: Retrieve all services owned by the calling user
```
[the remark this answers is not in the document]
giuseppedipinto: Not really, services are related to the UserSubscription, so they are somehow connected to a user.
However, we can improve the description

openapi.yaml
```diff
@@ -0,0 +1,730 @@
...
+          description: Number of services per page
+          required: true
+          schema:
+            type: number
+        - name: offset
+          in: query
+          description: Offset services number
+          required: false
+          schema:
+            type: number
```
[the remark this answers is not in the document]
giuseppedipinto: It depends, because if we retrieve services starting from subscriptions, Azure Apim has `$top` and `$skip` parameters and returns total `count` of retrieved subscriptions.

openapi.yaml
giuseppedipinto: i will remove the "|", but for the description I prefer the current version

openapi.yaml
```diff
@@ -0,0 +1,730 @@
...
+        '500':
+          description: Internal server error
+      security:
+        - api_key: []
+  /services/{serviceId}:
+    get:
+      tags:
+        - services
+      summary: Retrieve service
+      description: Retrieve a service by `serviceId`
```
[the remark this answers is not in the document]
giuseppedipinto: Same as above

openapi.yaml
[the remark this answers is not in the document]
giuseppedipinto: What do you think about using the `423 Locked` code to distinguish a really not found service _(404)_ from a canceled one _(423)_?

> The HTTP 423 Locked error response code indicates that either the resources tentatively targeted by is locked, meaning it can't be accessed.

openapi.yaml
```diff
@@ -0,0 +1,730 @@
...
+          description: Internal server error
+      security:
+        - api_key: []
+    delete:
+      tags:
+        - services
+      summary: Delete service
+      description: |
+        Delete a service by `serviceId`.<br>_Note: This API operation needs the __Manage
+        API key__ owned by the service user, otherwise `403 forbidden` will be returned to the caller._
```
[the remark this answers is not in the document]
rodoviti: @giuseppedipinto I just wanted to avoid referring to parameter names in the operation description. It's just a decoupling from what (description) and how (parameter, request/response body, ecc.) the operation does. To me, description is more a high level concept.
E.g.: we might have to change path param name from  _serviceId_ to _id_ and, with this "coupling", we will have to change also the operation description even if nothing has changed from that "side" (is not a change in behavior). I hope this example clarifies what i mean.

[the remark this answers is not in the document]
giuseppedipinto: @rodoviti you have been very clear and I don't see any problems in changing the form of the description, the substance does not change but as you rightly say, let's avoid direct references to parameters.
Let's say you almost convinced me! :-)

openapi.yaml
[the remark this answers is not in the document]
balanza: Let's reason agnostically: what are the common pagination patterns? I see:
1. `offset` and `limit`
2. `page` and `page_size`
3. `next_id` (provide the id of the next document to be retrieved)

We can usually make `1` out of `2` and vice versa. `3` is tricky as it's often very bound to the database query language.

openapi.yaml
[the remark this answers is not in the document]
rodoviti: mh, then I don't think it fits (at least in my head) what @balanza said regarding Authz 😄 (use a custom apiKey as a "machine to machine" auth)

openapi.yaml
giuseppedipinto: ac63b0f

openapi.yaml
```diff
@@ -441,6 +443,39 @@ paths:
...
+            type: string
+      responses:
+        '200':
+          description: Service history fetched successfully
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Service'
```
rodoviti, line 468: I think it needs to be "paginated"

openapi.yaml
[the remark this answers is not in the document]
giuseppedipinto: 2e38485e02312f20ce473591fc854b1e11de2913

openapi.yaml
giuseppedipinto: Correct, and by the way, I'll take the opportunity to add the `last_update` property in the `Service` model

giuseppedipinto: 8ff4109

rodoviti: if we use cosmos there already exists the out-of-the-box _\_ts_ field wich contains the last updated timestamp (but is at a second level of granularity) and [is already indexed by default](https://learn.microsoft.com/en-us/azure/cosmos-db/index-policy#includeexclude-strategy). So, theoretically, we can efficiently query by partitionKey (_serviceId_) sorted by updateDate (_\_ts_), but remains the offset "problem" of the pagination

giuseppedipinto: You're right, with CosmosDB we'll have only the `continuation` token concept to implement pagination...we need to discuss it later

openapi.yaml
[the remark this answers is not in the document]
giuseppedipinto: **_Final decision:_** `getService` of a deleted item will return `200` and the related service with `status=deleted`

StefanoDoc (approved):

PR #3, opened 2023-04-05
[IOPAE-253] Propose type interfaces for a Finite State Machine using `fp-ts`  

balanza:
#### List of Changes

* Scaffolding for the new package `@pagopa/fp-ts-fsm`
* Proposal of new type interfaces for FSM

#### Motivation and Context

Propose a type interface for a strict Finite State Machine definition. Actual implementations of FSM are not in the scope of this PR.

#### Checklist:


balanza:
Turn into the draft as we cannot build a generic type interface for a FSM so far.

PR #17, opened 2023-04-24
Refactor: `webservice` module

balanza:
#### List of Changes

See commit list

#### Motivation and Context

Isolate the web service module into a dedicated folder so that:
* every concept and resource related to http connection can be shared across all endpoints
* the entire app does not have to know about http interactions
* the webservice can be tested autonomously
* the webservice can be moved into another application if needed
* `main` is only responsible to mount the webservice to httpTrigger functions

#### Checklist:

### DIFF
--- a/packages/io-services-cms-models/package.json
+++ b/packages/io-services-cms-models/package.json
@@ -1,7 +1,9 @@
 {
   "name": "io-services-cms-models",
   "version": "0.0.0",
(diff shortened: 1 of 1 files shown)


StefanoDoc (approved):

ankit-gautam23 (commented):
LGTM

PR #20, opened 2023-04-27
[IOPAE-301] client apim config

StefanoDoc:
Add Apim configuration
#### List of Changes
Add config for apim plus some minor fixes

unit test

#### Checklist:


apps/io-services-cms-webapp/src/config.ts
```diff
@@ -32,12 +32,30 @@ export const JiraConfig = t.interface({
 });
 export type JiraConfig = t.TypeOf<typeof JiraConfig>;
 
-// global app configuration
+// Service Principal Creds configuration
+export const ServicePrincipalConfig = t.interface({
```
balanza, line 36: Naming: _ServicePrincipal_ of what? Also: we can use comments to explain **why** we write something, just repeating the code in plain English has usually little value.

apps/io-services-cms-webapp/src/jira_client.ts
```diff
@@ -18,24 +18,25 @@ export const CreateJiraIssueResponse = t.interface({
...
-          comment: t.interface({
-            comments: t.readonlyArray(t.interface({ body: t.string })),
-          }),
-          status: t.interface({
-            name: t.string,
-          }),
-        }),
-      })
-    ),
+    issues: t.readonlyArray(JiraIssue),
```
balanza, line 39: 👍

balanza (commented):
Should we add configs to `infra` as well?

apps/io-services-cms-webapp/src/config.ts
gunzip: FYI https://github.com/pagopa/developer-portal/pull/46#discussion_r1179390139

PR #21, opened 2023-04-28
[#IOPAE-302] APIM Util - Get Delegate from ServiceId

giuseppedipinto:
APIM Utility for retrieving service delegate information.

#### List of Changes

- Added _apim_ utility with `getDelegateFromServiceId` method
- Fix _apim_client_ `getUserGroups` error type return

#### Motivation and Context
We need to retrieve service owner data _(name, email, apim permissions)_ starting from `serviceId`.

Manual integration test

#### Checklist:


apps/io-services-cms-webapp/src/utils/apim.ts
```diff
@@ -0,0 +1,46 @@
...
+              apim,
+              user.name as NonEmptyString
+            ),
+            TE.map(
+              (userGroups) =>
+                ({
+                  name: `${user.firstName} ${user.lastName}`,
+                  email: user.email,
+                  permissions: userGroups.map((group) => group.name),
+                } as Delegate)
```
balanza, line 40: Can we avoid manual casting? This could lead to an issue if we update the `Delegate` definition in the future.
Consider these alternatives:
* make the object to be in the shape of a `Delegate` (so that no cast is required)
* decode the object (to enforce strict validation of data)
* relax the `Delegate` definition (if you think it's unnecessarily strict and unhandy to use)

balanza (changes requested):

balanza (approved):

PR #47, opened 2023-05-09
[IOPAE-311] kebab-case renaming

giuseppedipinto:
Project renaming with `kebab-case` convenction.

#### Checklist:


balanza (approved):

StefanoDoc (approved):

balanza:
This refactoring led to the discovery of the bug explained in https://github.com/pagopa/io-services-cms/pull/48. 

> TL;DR there is an unreferenced file we cannot remove; otherwise, the build fails. 

We decided to continue with the job anyway, and just leave such a file until a solution has been found. The file has been renamed into `_to_solve.ts`.

PR #50, opened 2023-05-11
feat: [IOPAE-315] setup timerTrigger fb for service review checker (poller)

rodoviti:
Create a timeTrigger function and its entryPoint in order to check the status to all pending reviews

#### List of Changes

- add timerTrigger config
- add timerTrigger entryPoint

#### Motivation and Context

To check all Jira ticket updates related to all pending review

no test for "config" code

#### Checklist:


balanza (changes requested):
* we need to configure `SERVICE_REVIEW_CHECKER_SCHEDULE` otherwise the app would break
* consider disable the handler so that it doesn't run until we are ready
```
"AzureWebJobs.ServiceReviewChecker.Disabled" = "1"
```

giuseppedipinto (approved):
LGTM

apps/io-services-cms-webapp/ServiceReviewChecker/function.json
```diff
@@ -0,0 +1,13 @@
+{
+  "disabled": true,
```
balanza, line 2: Unfortunately, this flag doesn't work (although docs are not clear about it). The safest method to disable the function is to add this to application settings:
```
"AzureWebJobs.ServiceReviewChecker.Disabled" = "1"
```

balanza (changes requested):

balanza (approved):
I've added and applied this configuration to disable the function to start: https://github.com/pagopa/io-services-cms/pull/50/commits/5735c341367d2624af22fdaf9696462f9a19d409

We can merge and hence publish when we are ready

PR #53, opened 2023-05-16
[IOPAE-318] Setup `OnLegacyServiceChange` function

balanza:
#### List of Changes

* refactor: split `webapp` file to increase readability in https://github.com/pagopa/io-services-cms/commit/8b7378072e0e1e68d005cd25049032dddd0f2c5c
* set connection to legacy cosmos db in  https://github.com/pagopa/io-services-cms/commit/70368896dcded00595368310aa6e2acc94a0923b
* scaffold a cosmosdb-triggered function in https://github.com/pagopa/io-services-cms/commit/b98b3e8b47b4d05fe7c48cbf00e58daf0122dc99

#### Motivation and Context

We want to import current services into our new model. To do so we connect to the relative change feed and map every change to every service into a state of our FSM for both `ServicesLifecycles` and `ServicesPublication`.
The handler for the change feed is initially disabled 

#### Checklist:


infra/src/webapp.tf
```diff
@@ -225,7 +236,12 @@ module "webapp_functions_app_staging_slot" {
 
   always_on = "true"
 
-  app_settings = local.webapp_functions_app_settings
+  app_settings = merge(
+    local.webapp_functions_app_settings,
+    {
+      "AzureWebJobs.OnLegacyServiceChange.Disabled" = "1"
+    }
+  )
```
balanza, line 244: Please note that this configuration is explicitly written both in the production and staging slots. This is because **listeners must be kept disabled in the staging slot, otherwise they will _steal_ items from queues/feeds**, being them listening to the same resources of production.

Alternatively, we can use dedicated queues/feeds for staging slots (that could be interesting).

giuseppedipinto: Ok clear, at the moment I would keep this double disabling

giuseppedipinto (approved):
LGTM

PR #54, opened 2023-05-16
[IOPAE-256] Service Publication FSM

giuseppedipinto:
#### List of Changes

* Define the FSM and implement it
* Define common types for FMS

#### Motivation and Context

Implements a finite state machine to handle the publication of a service as described in the following diagram:
```mermaid
stateDiagram-v2
    * --> unpublished: override
    unpublished --> unpublished: override
    published --> published: override
    unpublished --> published: publish
    published --> unpublished: unpublish
```

See [README](https://github.com/pagopa/io-services-cms/blob/IOPAE-256--fsm-service-publication/packages/io-services-cms-models/README.md) for further documentation.

With a local store.

#### Checklist:

### DIFF
--- a/.changeset/empty-toys-wave.md
+++ b/.changeset/empty-toys-wave.md
@@ -1,0 +1,5 @@
+---
+"@io-services-cms/models": minor
+---
+
+Service Publication FSM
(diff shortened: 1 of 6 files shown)


rodoviti (approved):
LGTM,  but I didn't really understand the initial state transition (from "void" to unpublish).

PS: I only focused on FSM transitions

PR #55, opened 2023-05-16
[IOPAE-323] FSM Custom Errors

giuseppedipinto:
Add **FSM custom Errors** instead of generic Error only:
- `FsmNoApplicableTransitionError` _(in case of no transition declared for an action)_
- `FsmNoTransitionMatchedError` _(in case of no transition that matches with the specified appliedAction)_
- `FsmTooManyTransitionsError` _(avoid indeterministic situations in which there are more than one transition applicable)_
- `FsmTransitionExecutionError` _(in case of error on transition execution)_
- `FsmStoreFetchError` _(in case of store fetch operation error)_
- `FsmStoreSaveError` _(in case of store save operation error)_

Unit tests

#### Checklist:


balanza:
I was not sure about this solution even if I sponsored it, so I played a bit [in a forked branch](https://github.com/pagopa/io-services-cms/blob/IOPAE-323-fsm-errors-enhancement-2), and I tried the following:

**If the signature of `apply` matched what it's actually returned**
[here](https://github.com/pagopa/io-services-cms/blob/IOPAE-323-fsm-errors-enhancement-2/packages/io-services-cms-models/src/service-lifecycle/index.ts#L34) I removed some error kinds and I expected the build to fail, as we are returning kinds not defined in signature. It doesn't, meaning we cannot trust the signature.

This happens because Typescript doesn't evaluate types by their name (like Java or C#), it evaluates the structure. So two classes with different names but the same fields, end up being the same type for Typescript.

We must differentiate objects somehow. One possible solution is to add a `kind` field to each class, like:
```ts
class FsmNoTransitionMatchedError extends Error {
  public kind = "FsmNoTransitionMatchedError";
  constructor() {
    super(`No transition matched`);
  }
}
```

Apart from that, we still have: 
* We can narrow types correctly (as proved [here](https://github.com/pagopa/io-services-cms/blob/01084af867993ef8e11ef311405e332c28c89757/packages/io-services-cms-models/src/service-lifecycle/__tests__/index.test.ts#L129))
* We can read inform accurately the kind of error that occurred 

...and that is very cool 😎

balanza (approved):
Great job 🚀

balanza:
The `apply` function is merely a mechanism that should be agnostic about any specific FSM definition. However, we couldn't make it a commodity for every FSM, yet. This is mainly because we failed to provide an FSM-specific signature while keeping `apply` a general-purpose function.

Each FSM should expose its own `apply` signature depending on actual defined states/actions/transitions and how they relate. So far, we only achieved to express such specificity using hand-written overloads.

So, we decided to duplicate the `apply` function implementation for each FSM waiting for a more general solution.

PR #56, opened 2023-05-17
feat: [IOPAE-312] add on service lifecycle change handler

rodoviti:
add a handler to create all "actions" to be applied to each "changed" item.

#### List of Changes

- implement a new common type RequestReviewItem into packeges/model
- implement "single item" handler
- implement unit tests

#### Motivation and Context

This change is required in order to manage (actually, to properly dispatch) every ServiceLifecycle changes

unit tests

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/env.example
+++ b/apps/io-services-cms-webapp/env.example
@@ -1,5 +1,7 @@
 # Please refer to src/config.ts for the actual configuration
 
+AzureWebJobsStorage=DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://fnstorage:10000/devstoreaccount1;QueueEndpoint=http://fnstorage:10001/devstoreaccount1;TableEndpoint=http://fnstorage:10002/devstoreaccount1;
+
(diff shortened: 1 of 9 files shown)


apps/io-services-cms-webapp/env.example
```diff
@@ -32,4 +32,14 @@ AZURE_APIM=
 AZURE_APIM_RESOURCE_GROUP= 
 AZURE_SUBSCRIPTION_ID=
 
-AzureWebJobsStorage=DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://fnstorage:10000/devstoreaccount1;QueueEndpoint=http://fnstorage:10001/devstoreaccount1;TableEndpoint=http://fnstorage:10002/devstoreaccount1;
```
giuseppedipinto, line 35: Why this parameter has been removed?

rodoviti: There is no reference in either the `config.ts` or the `webapp.tf` file, so I assumed it was an unused value.
I think it is a configuration for the old version of the _tf_ module... @balanza can you confirm that?

balanza:
(sorry, I closed the PR by mistake)

apps/io-services-cms-webapp/env.example
balanza: It's a mandatory value as it's expected by Azure Function ([docs](https://learn.microsoft.com/en-my/azure/azure-functions/functions-app-settings)).

We can decide to enforce validation by adding it to `config.ts`. We don't always do it as we take the value for granted. I think the app won't even start without it (I don't remember right now)

giuseppedipinto (changes requested):
Please add `AzureWebJobsStorage` according to @balanza explanation

apps/io-services-cms-webapp/env.example
rodoviti: OK...so, the fact that it is not in the `webapp.tf` file (as an env var) is an error, right?

giuseppedipinto (approved):

PR #61, opened 2023-05-18
bugfix: [IOPAE-325] add review service fn

rodoviti:
add review service fn

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/ReviewService/function.json
+++ b/apps/io-services-cms-webapp/ReviewService/function.json
@@ -1,0 +1,21 @@
+{
+    "bindings": [
+        {
+            "authLevel": "anonymous",
+            "type": "httpTrigger",
+            "direction": "in",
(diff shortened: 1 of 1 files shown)


apps/io-services-cms-webapp/ReviewService/function.json
```diff
@@ -0,0 +1,21 @@
+{
+    "bindings": [
+        {
+            "authLevel": "anonymous",
+            "type": "httpTrigger",
+            "direction": "in",
+            "name": "req",
+            "route": "services/:serviceId/review",
```
giuseppedipinto, line 8: it should be with curly braces:
```suggestion
            "route": "services/{service_id}/review",
```

giuseppedipinto (changes requested):

giuseppedipinto (approved):

PR #62, opened 2023-05-18
fix: storage account public access unless private endpoint

manuelcoppotelli:
A Storage Account requires either public access or a private endpoint. Otherwise its endpoint cannot be accessed and no operation on it can be performed

#### Checklist:


balanza (approved):
thanks @manuelcoppotelli, we appreciate a lot!

PR #82, opened 2023-05-24
[#IOPAE-250] Make approved service as publicable

giuseppedipinto:
Mapping a service **approve** action on _ServiceLifecycle FSM_ to **override** action on _ServicePublication FSM_
#### List of Changes
- Add new queue output item on ServiceLifecycleWatcher
- Add new OnRequestPublication queueTrigger function triggered on requestPublication items
- Update _approved_ unit test scenario

#### Checklist:


apps/io-services-cms-webapp/src/watchers/on-services-lifecycles-change.ts
```diff
@@ -3,13 +3,20 @@ import * as RTE from "fp-ts/lib/ReaderTaskEither";
 import * as TE from "fp-ts/lib/TaskEither";
 import { pipe } from "fp-ts/lib/function";
 
-type Actions = "requestReview" | "other";
+type Actions = "requestReview" | "requestPublication" | "other";
```
rodoviti, line 6:
```suggestion
type Actions = "requestReview" | "requestPublication";
```

apps/io-services-cms-webapp/OnRequestPublication/function.json
```diff
@@ -0,0 +1,13 @@
+{
+  "bindings": [
+    {
+      "queueName": "request-publication",
```
rodoviti, line 4:
```suggestion
      "queueName": "%REQUEST_PUBLICATION_QUEUE%",
```

rodoviti: @giuseppedipinto can you please do the same change to `OnRequestReview/function.json` (i forgot about that)? Thanks

PS: if you want to add more control, you can check for the existence of these variables in config.ts 😉

apps/io-services-cms-webapp/src/publicator/request-publication-handler.ts
```diff
@@ -0,0 +1,47 @@
...
+    ServiceLifecycle.definitions.Service.decode,
+    E.mapLeft(flow(readableReport, (_) => new Error(_)))
+  );
+
+export class RequestPublicationParseError extends Error {
+  public kind = "RequestPublicationParseError";
+  constructor() {
+    super(`Error while parsing incoming message`);
+  }
+}
```
rodoviti, line 27: IMHO there is no need to create a custom error, use just  a `new Error("Error while parsing incoming message")`

apps/io-services-cms-webapp/src/publicator/request-publication-handler.ts
```diff
@@ -0,0 +1,47 @@
...
+      TE.fromEither,
+      TE.mapLeft((_) => new RequestPublicationParseError()),
+      TE.chain((service) =>
+        ServicePublication.apply("override", service.id, { data: service })(
+          store
+        )
+      ),
+      TE.getOrElse((e) => {
+        throw e;
+      })
```
rodoviti, line 45: will the error be logged automatically by the az fn? if don't, we have to at lest log it 😉 

PS: @balanza is it possible to add a new field to item before it will be stored to poison queue? I am thinking of storing the error (message, type, or other) to have enough data to analyze the problem directly from the poison queue.

rodoviti (approved):
LGTM except for those little changes that I suggested 😉

apps/io-services-cms-webapp/src/publicator/request-publication-handler.ts
giuseppedipinto: I'm pretty sure that _function throw_ is automatically logged by azure.

apps/io-services-cms-webapp/src/watchers/on-services-lifecycles-change.ts
giuseppedipinto: c00ca73

apps/io-services-cms-webapp/OnRequestPublication/function.json
giuseppedipinto: c00ca73

apps/io-services-cms-webapp/src/publicator/request-publication-handler.ts
giuseppedipinto: c00ca73

PR #99, opened 2023-06-07
[#IOPAE-340] set max_allowed_payment_amount on publication

StefanoDoc:
we want publish service to be able to send messages with attached payment notices , so we need to automatically set the **max_allowed_payment_amount** field to 1000000 when the services is published
#### List of Changes
on-services-lifecycle-change -> onApproveHandler -> requestPublication -> we set max_allowed_payment_amount

tested with a local execution of the app

#### Checklist:


apps/io-services-cms-webapp/src/watchers/__tests__/on-services-lifecycles-change.test.ts
```diff
@@ -24,12 +24,33 @@ const aService = {
   },
 } as unknown as ServiceLifecycle.ItemType;
 
+const aPublicationService = {
```
giuseppedipinto, line 27: Try to avoid service complete duplication using the spread operator, for example:
`const aPublicationService = {...aService, max_allowed_payment_amount: 1000000}`

StefanoDoc: i tried but it doesn't work like that , it adds max_allowed_payment_amount like a new parameter at the end of the service , and does not change the one inside service.data.max_allowed_payment_amount

giuseppedipinto (approved):

rodoviti (approved):

rodoviti:
@StefanoDoc remember to add the changeset before merge 😉

PR #100, opened 2023-06-07
[IOPAE-260] Add automatic publication capability on service review API

giamma1295:
Added autoPublish capability, now when a service is submitted to be reviewed can be requested to be automatically published on review approval.
#### List of Changes
- ADD body payload with _auto_\__publish_ flag on API PUT /services/:serviceId/review
- ADD args:{autoPublish: boolean} params on FSM submit action
- ADD new publish action on FSM publication, this action will allow a service to be published form state _*_

Unit Test and Local application run

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/webservice/controllers/__test__/edit-service.test.ts
+++ b/apps/io-services-cms-webapp/src/webservice/controllers/__test__/edit-service.test.ts
@@ -41,7 +41,7 @@ describe("editService", () => {
       department_name: "string",
     },
     require_secure_channel: true,
-    authorized_recipients: ["12345678901"],
+    authorized_recipients: ["AAAAAA00A00A000A"],
     max_allowed_payment_amount: 0,
     metadata: {
       web_url: "string",


apps/io-services-cms-webapp/src/publicator/__test__/request-publication-handler.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      {} as unknown as ServicePublication.FsmClient;
+
+    try {
+      await handleQueueItem(
+        context,
+        anInvalidQueueItem,
+        mockFsmLifecycleClient
+      )();
+    } catch (error) {
+      expect(error).toHaveProperty("message");
```
gunzip, line 58: this test will pass even if the exception is not throw. you need one between:

A) declare the number of expectations using [expect.assertions](https://jestjs.io/docs/expect#expectassertionsnumber)

B) call fail() after `handleQueueItem`

apps/io-services-cms-webapp/src/publicator/request-publication-handler.ts
```diff
@@ -1,34 +1,54 @@
...
+          })
+        : fsmPublicationClient.override(service.id, {
+            data: {
+              id: service.id,
+              data: service.data,
+            },
+          })
+    ),
+    TE.getOrElse((e) => {
+      throw e;
```
gunzip, line 45: I guess this `throw` causes a retry in the function handler. You probably want to tell transient errors from permament ones (ie. if the format of the payload is wrong, causing n-retries here would consume costly cycles without a reason).

apps/io-services-cms-webapp/src/publicator/__test__/request-publication-handler.test.ts
giamma1295: Thanks for the suggestion, the test has been modified to fail in case the decoding of the queued message does not fail using the following vitest components:

- [expect.rejects](https://vitest.dev/api/expect.html#rejects)
- [toThrowError](https://vitest.dev/api/expect.html#tothrowerror)

```typescript
await expect(() =>
      handleQueueItem(context, anInvalidQueueItem, mockFsmLifecycleClient)()
    ).rejects.toThrowError("Error while parsing incoming message");
```

apps/io-services-cms-webapp/src/publicator/request-publication-handler.ts
giamma1295: We will handle this issue with new pull request in the near future

gunzip: in these cases add a comment with a TODO and a reference to the JIRA ID  of the issues

apps/io-services-cms-webapp/src/publicator/request-publication-handler.ts
```diff
@@ -1,34 +1,54 @@
...
+              data: service.data,
+            },
+          })
+        : fsmPublicationClient.override(service.id, {
+            data: {
+              id: service.id,
+              data: service.data,
+            },
+          })
+    ),
```
rodoviti, line 43:
```suggestion
    TE.chainW((service) =>
      pipe(
        {
          data: {
            id: service.id,
            data: service.data,
          },
        },
        (publicationArgs) =>
          service.autoPublish
            ? fsmPublicationClient.publish(
                publicationArgs.data.id,
                publicationArgs
              )
            : fsmPublicationClient.override(service.id, publicationArgs)
      )
    ),
```

apps/io-services-cms-webapp/src/watchers/on-services-lifecycles-change.ts
```diff
@@ -34,7 +34,11 @@ const onSubmitHandler = (
 const onApproveHandler = (
   item: ServiceLifecycle.ItemType
 ): RequestPublicationAction => ({
-  requestPublication: { id: item.id, data: item.data },
+  requestPublication: {
+    id: item.id,
+    data: item.data,
+    autoPublish: (item.fsm.autoPublish as boolean) ?? false,
```
rodoviti, line 40:
```suggestion
    autoPublish: (item.fsm.autoPublish as boolean) ?? false, // if FSM Lifecycle is in _approved_ state, then `autopublish` will always be set
```

packages/io-services-cms-models/src/service-lifecycle/index.ts
```diff
@@ -164,6 +168,7 @@ const FSM: FSM = {
           fsm: {
             state: "approved",
             approvalDate,
+            autoPublish: current.fsm.autoPublish ?? false,
```
rodoviti, line 171: if is possibile, try to apply DRY principles....we avoid repeating `autopublish` default value 😉

apps/io-services-cms-webapp/src/watchers/on-services-lifecycles-change.ts
rodoviti: if is possibile, try to apply DRY principles....we avoid repeating `autopublish` default value 😉

rodoviti (changes requested):
see suggested changes.

PS: why openapi.yaml fromat is changed? it seems you have a different formatter/prettier configuration than ours...

apps/io-services-cms-webapp/src/publicator/request-publication-handler.ts
```diff
@@ -1,34 +1,54 @@
...
+export const handleQueueItem = (
+  context: Context,
+  queueItem: Json,
+  fsmPublicationClient: ServicePublication.FsmClient
+) =>
+  pipe(
+    queueItem,
+    parseIncomingMessage,
+    TE.fromEither,
+    TE.mapLeft((_) => new Error("Error while parsing incoming message")),
```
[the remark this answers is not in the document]
giamma1295: Accepted

apps/io-services-cms-webapp/src/watchers/on-services-lifecycles-change.ts
giamma1295: We added a function `getAutoPublishm` in order to handle the default value for `autoPublish` globally.

packages/io-services-cms-models/src/service-lifecycle/index.ts
giamma1295: We added a function `getAutoPublish` in order to handle the default value for `autoPublish` globally.

iwoak (approved):
LGTM

PR #132, opened 2023-06-15
[#IOPAE-350] Update GET Services openapi query params

giuseppedipinto:
Update **GET Services** openapi query params:
- change `limit` to optional, update description and define range values _(1-100)_ and default value _(20)_
- update `offset` description and define minimum and default value _(0)_

#### Checklist:


rodoviti (approved):
nitpick: should these changes be combined with other changes (i.e.: controller handler/middleware)?

PR #137, opened 2023-06-19
[#IOPEA-369] Fix withDefault config keys

giuseppedipinto:
Fix webapp configuration keys in which a default value is expected.
With this fix it is possible to use correctly the default value of the "_withDefault_" keys without necessarily having to enter the value in the environment.

#### List of Changes
- apps/io-services-cms-webapp/src/config.ts

#### Checklist:


rodoviti (approved):
LGTM

PR #139, opened 2023-06-19
[#IOPAE-363] Add Manage Key Middleware

giamma1295:
#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/utils/cosmos-legacy.ts
+++ b/apps/io-services-cms-webapp/src/utils/cosmos-legacy.ts
@@ -1,0 +1,20 @@
+/**
+ * Use a singleton CosmosDB client across functions.
+ */
+import { CosmosClient } from "@azure/cosmos";
+
+import { getConfigOrThrow } from "../config";
+
+const config = getConfigOrThrow();
+
+// Setup DocumentDB
+export const cosmosDbUri = config.COSMOSDB_LEGACY_URI;
+export const cosmosDbName = config.COSMOSDB_LEGACY_NAME;
+export const cosmosDbKey = config.COSMOSDB_LEGACY_KEY;
+
+export const cosmosdbClient = new CosmosClient({
+  endpoint: cosmosDbUri,
+  key: cosmosDbKey,
+});
+
+export const cosmosdbInstance = cosmosdbClient.database(cosmosDbName);


apps/io-services-cms-webapp/src/utils/subscription.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @param config
+ * @param apimClient
+ * @param serviceId
+ * @param ownerSubscriptionId subscriptionId related to 'Ocp-Apim-Subscription-Key'
+ * @param userId APIM userId
+ * @returns
+ */
+export const serviceOwnerCheckManageTask = (
+  config: IConfig,
```
gunzip, line 38: General advice: in order to minimize the number of dependencies (parameters) passed into a method, consider passing it all the needed dependencies already configured (in this case, the configured `getSubscription` method), ie:

```ts
const getSubscriptionT = (apimClient: ApiManagementClient, config: IConfig) => (serviceId: NonEmptyString) => {
  getSubscription(
      apimClient,
      config.AZURE_APIM_RESOURCE_GROUP,
      config.AZURE_APIM,
      serviceId
    );

export const serviceOwnerCheckManageTask = (getSubscription: typeof ReturnValue<getSubscriptionT>, serviceId: NonEmptyString, ownerSubscriptionId: NonEmptyString,
  userId: NonEmptyString) => { ... getSubscription(serviceId) ... }
```

the advantages of this approach are more evident when you combine ReaderTaskEither(s), in order to pass _all_ the dependencies only once when calling the outer `ReaderTaskEither`. In order to do this you have to invert parameters order:

```ts
interface GetSubscriptionEnv {
  apimClient: ApiManagementClient;
  config: IConfig;
}

const getSubscriptionRTE =  (serviceId: NonEmptyString): 
   RTE.ReaderTaskEither<GetSubscriptionEnv, Error, Subscription> => 
({apimClient, config}: GetSubscriptionEnv) => {
  getSubscription(
      apimClient,
      config.AZURE_APIM_RESOURCE_GROUP,
      config.AZURE_APIM,
      serviceId
    );
type GetSubscriptionRTE = typeof getSubscriptionRTE;

// the following method returns a ReaderTaskEither
export const serviceOwnerCheckManage  = (serviceId: NonEmptyString, ownerSubscriptionId: NonEmptyString,
       userId: NonEmptyString) => pipe(RTE.Do, ... getSubscriptionRTE(serviceId) ... )

// ... later, on caller ...

const doSomething = (serviceId: NonEmptyString, ownerSubscriptionId: NonEmptyString,
       userId: NonEmptyString) =>

pipe(
  RTE.Do,
....
  RTE.bind("something", serviceOwnerCheckManage(serviceId, ownerSubscriptionId, 
      userId)), ...

// here you pass all the needed dependencies once:
)( {  apimClient, config  } )


```

In this way you can delete all the `makeSomethingHandler` methods and use them directly as RTE.

(not needed for this PR anyway)

apps/io-services-cms-webapp/src/utils/subscription.ts
```diff
@@ -0,0 +1,64 @@
...
+      serviceId
+    ),
+    TE.mapLeft(
+      () =>
+        ResponseErrorInternal(
+          `An error has occurred while retrieving service '${serviceId}'`
+        ) // FIXME: handle the error correctly
+    ),
+    TE.chainW((serviceSubscription) =>
+      ownerSubscriptionId.startsWith("MANAGE-")
```
gunzip, line 58: If I understand well when `ownerSubscriptionId.startsWith("MANAGE-") === false` we don't need to call `getSubscription` API (which we do anyway), isn't it? You probably want to handle this case first. I'd give a name to this boolean check extracting it in its own method:

`const isManageKey = (ownerSubscriptionId: NonEmptyString) => ownerSubscriptionId.startsWith("MANAGE-");
`

giamma1295: Thanks for the suggested improvement, we have made the change and now we execute as the first thing in the pipe that the passed key is a manageKey.

``` typescript
...
  pipe(
    ownerSubscriptionId,
    TE.fromPredicate(
      (oSId: NonEmptyString) => isManageKey(oSId),
      () => ResponseErrorForbiddenNotAuthorized
    ),
    TE.chainW(() =>
      pipe(
        getSubscription(
          apimClient,
...
```

apps/io-services-cms-webapp/src/utils/subscription.ts
```diff
@@ -0,0 +1,74 @@
...
+  config: IConfig,
+  apimClient: ApiManagementClient,
+  serviceId: NonEmptyString,
+  ownerSubscriptionId: NonEmptyString,
+  userId: NonEmptyString
+): TaskEither<ErrorResponses, NonEmptyString> =>
+  pipe(
+    ownerSubscriptionId,
+    TE.fromPredicate(
+      (ownSubId: NonEmptyString) => isManageKey(ownSubId),
```
rodoviti, line 51: nitpick: syntactic sugar
```suggestion
      isManageKey,
```

apps/io-services-cms-webapp/src/utils/subscription.ts
```diff
@@ -0,0 +1,74 @@
...
+        TE.mapLeft(() =>
+          ResponseErrorInternal(
+            `An error has occurred while retrieving service '${serviceId}'`
+          )
+        )
+      )
+    ),
+    TE.chainW((serviceSubscription) =>
+      serviceSubscription.ownerId === userId
+        ? TE.of(serviceId)
```
rodoviti, line 71: **suggestion(non-blocking):** IMHO this way it's more readable 😉 
```suggestion
        ? TE.right(serviceId)
```
Otherwise you can use a TE.fromPredicate instead of ternary operator

rodoviti (approved):
LGTM

apps/io-services-cms-webapp/src/utils/subscription.ts
giamma1295: suggestion included in commit [suggestion code cleanUp](https://github.com/pagopa/io-services-cms/pull/139/commits/28f33caeae8a6690bf03b2c2483f459a56dafecd)

apps/io-services-cms-webapp/src/webservice/controllers/get-service-publication.ts
```diff
@@ -45,35 +56,51 @@ type HandlerResponseTypes =
...
+      TE.chainW(
+        flow(
+          store.fetch,
+          TE.mapLeft((err) => ResponseErrorInternal(err.message)),
+          TE.map((res) =>
+            O.isSome(res)
+              ? ResponseSuccessJson<ServiceResponsePayload>(
+                  itemToResponse(res.value)
+                )
+              : ResponseErrorNotFound("Not found", `${serviceId} not found`)
```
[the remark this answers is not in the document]
giamma1295: suggestion included in commit [suggestion code cleanUp](https://github.com/pagopa/io-services-cms/pull/139/commits/28f33caeae8a6690bf03b2c2483f459a56dafecd)

apps/io-services-cms-webapp/src/utils/subscription.ts
giamma1295: Thanks fro the tip, a ticket to make this improvement has been created

PR #140, opened 2023-06-19
feat: [IOPAE-320]  add force/override action on fsm

rodoviti:
Add override action to both FSM (lifecycle and publication) to bypass FSM constraints and "force" update when sync from legacy in order to get both system in-sync

#### List of Changes

- add override action to FSM lifecycle
- add override action to FSM publication

#### Motivation and Context

It is required to get in-synch both legacy and new cms systems

#### How Has This Been Tested?
Unit tests

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #149, opened 2023-06-21
Refactor on #140

balanza:
Just a refactor on https://github.com/pagopa/io-services-cms/pull/140.

The aim is to flatten a code with too many indentations.

### DIFF
--- a/packages/io-services-cms-models/src/service-lifecycle/index.ts
+++ b/packages/io-services-cms-models/src/service-lifecycle/index.ts
@@ -430,15 +430,12 @@ function override(
       store.fetch(id),
       TE.chain(
         flow(
-          O.fold(
-            () => E.right(void 0),
-            flow(
-              ItemType.decode,
-              E.bimap(
-                flow(readableReport, (msg) => new FsmItemValidationError(msg)),
-                (_) => void 0
-              )
-            )
+          // if we found an item, we must validate by decode with ItemType
+          O.map(ItemType.decode),
+          // else, if no item is found we just assume it's ok
+          O.getOrElseW(() => E.right(void 0)),
+          E.mapLeft(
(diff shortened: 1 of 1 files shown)


rodoviti (approved):
thanks 😉

packages/io-services-cms-models/src/service-lifecycle/index.ts
```diff
@@ -430,15 +430,12 @@ function override(
...
-            )
+          // if we found an item, we must validate by decode with ItemType
+          O.map(ItemType.decode),
+          // else, if no item is found we just assume it's ok
+          O.getOrElseW(() => E.right(void 0)),
+          E.mapLeft(
+            flow(readableReport, (msg) => new FsmItemValidationError(msg))
           ),
           TE.fromEither
         )
```
balanza, line 441: **issue:** ‏What is the value of fetching and validating what it's *currently* in the store? I think the value to the user would be to validate what's going to be inserted, hence preventing bad data to pollute the database. the storage 

I also see this possible issue: what if we _somehow_ stored malformed data, maybe because of a bug? We won't be able to override it, as the validation will fail and so the override operation.

rodoviti: I thought we need to validate our input on the "boundary" of our application, and then rely on strong typing. This is not an "application boundary", isn't it?

The validation after fetching from store actually doesn't make much sense...

balanza: > The validation after fetching from store actually doesn't make much sense

My point. The assumption of the `override` function would be:

> I let you store an arbitrary object in the database as long as it is well-formed, meaning:
> * it has a state defined in the FSM
> * it has a shape consistent with what's meant for that state in the FSM

Maybe just the function signature will make it.

PR #154, opened 2023-06-22
[#IOPAE-376] API Regenerate service keys

giuseppedipinto:
Add new API for regenerate service keys.

#### List of Changes
- Added new _**regenerateSubscriptionKey**_ functionality on _apim-client_
- Added new _**RegenerateServiceKeys**_ AZ Function and handler
- Added related _unit tests_

#### Motivation and Context
We need to regenerate security keys for a given service.

Unit tests

#### Checklist:


apps/io-services-cms-webapp/RegenerateServiceKeys/function.json
```diff
@@ -0,0 +1,21 @@
+{
+    "bindings": [
+        {
+            "authLevel": "anonymous",
```
gunzip, line 4: are these function being called passing through the API gateway? I'm wondering if is there any reason on keeping them accessible without an host key (anonymous access)

giuseppedipinto: The reason is that we have not addressed this issue yet _(we will address it in the next few days)_ and we will do it for all functions.

apps/io-services-cms-webapp/src/webservice/controllers/__test__/regenerate-service-keys.test.ts
```diff
@@ -0,0 +1,277 @@
...
+const aManageSubscriptionId = "MANAGE-123";
+const ownerId = "123";
+const primaryKey = "any-primary-key-value";
+const secondaryKey = "any-secondary-key-value";
+
+const anApimResource = {
+  id: "any-id",
+  name: "any-name",
+  _etag: "_etag",
+  ownerId,
```
giamma1295, line 52: When apim is called the returned object as the property `ownerId` valued in this format "something/.../userId", is this handled correctly?

giamma1295, line 52: When apim is called the returned object as the property ownerId valued in this format "something/.../userId", is this handled correctly?

giamma1295 (changes requested):
Aside from the possible ownerId problem, this pr LGTM, so just check if the ownerId is handled correctly.

apps/io-services-cms-webapp/src/webservice/controllers/__test__/regenerate-service-keys.test.ts
giuseppedipinto: thanks for reporting, I checked it and the format is not relevant for the purpose of these unit tests ;-)

giamma1295 (approved):

PR #155, opened 2023-06-22
[#IOPAE-359] ServiceReviewChecker CRON Expression

giuseppedipinto:
Definition of the _cron expression_ of the **ServiceReviewChecker** timerTriggered function.
The new cron `0 9-18 * * 1-5` stands for **"Every hour, between 09:00 AM and 06:59 PM, Monday through Friday".**

#### List of Changes
- io-services-cms/apps/io-services-cms-webapp/ServiceReviewChecker/function.json

#### Checklist:


apps/io-services-cms-webapp/ServiceReviewChecker/function.json
```diff
@@ -1,12 +1,12 @@
 {
   "bindings": [
     {
-      "schedule": "* * * * *",
+      "schedule": "0 9-18 * * 1-5",
```
rodoviti, line 4: **suggestion(non-blocking):** I would recommend staying broader with the time window.
Is the time local or not?
```suggestion
      "schedule": "0 8-19 * * 1-5",
```

rodoviti (approved):

PR #156, opened 2023-06-22
feat: [IOPAE-382]  sync legacy watcher

rodoviti:
Create LegacyServiceWatcher
#### List of Changes

- add OnLegacyServiceChangeEntryPoint
- add OnLegacyServiceChangeHandler

#### Motivation and Context

This watcher is required to keep in sync changes from Legacy system to the new CMS

#### Checklist:


apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
```diff
@@ -1,3 +1,48 @@
...
+    return {} as ServiceLifecycle.ItemType;
+  }
+};
+
+export const handler: RTE.ReaderTaskEither<
+  { item: LegacyService },
+  Error,
+  NoAction | RequestSyncCmsAction
+> = ({ item }) => {
+  if (!!item.csmTag && item.csmTag) {
```
gunzip, line 43: is there any value for `item` where `item.csmTag` is `false` but `!!item.csmTag` is `true`?

apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
```diff
@@ -1,3 +1,48 @@
...
+const legacyToCms = (item: Service): Queue.RequestSyncCmsItem => {
+  // TODO: implement mapping
+  if (item.isVisible) {
+    return {} as ServicePublication.ItemType;
+  } else {
+    return {} as ServiceLifecycle.ItemType;
+  }
+};
+
+export const handler: RTE.ReaderTaskEither<
```
gunzip, line 38: if this method is a stub (it returns a TaskEither even if it is synchronous) add it to some comment in order to make reviews easier.

apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
rodoviti: actually, no. This is only an example/placeholder because we have not yet decided either the name or the type of the field (it could be either a boolean or an "enum like" type)

apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
rodoviti: good point, thsnks! 😉

rodoviti: fix with d2911a67c09effce37e41cc79709b163e286e15f

apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
rodoviti: fix with d2911a67c09effce37e41cc79709b163e286e15f

StefanoDoc (approved):

PR #158, opened 2023-06-23
fix: owner id match and wrong cosmosDb legacy name

giamma1295:
When apim get.subscrition is called the returned object has the property `ownerId` format is the following `/somenthing/.../userId`, this was causing an issue while checking for user service ownership.
#### List of Changes

- add `extractOwnerId` method in `subscription.ts` which will extract the userId from apim response.
- fix typo on cosmosDB legacy name

unit tests

#### Checklist:


apps/io-services-cms-webapp/src/utils/subscription.ts
```diff
@@ -24,6 +29,23 @@ type ErrorResponses =
 const isManageKey = (ownerSubscriptionId: NonEmptyString) =>
   ownerSubscriptionId.startsWith(MANAGE_APIKEY_PREFIX);
 
+const areUserAndOwnerEquals = (
```
gunzip, line 32: do we need really need a method just to call equal? :D

apps/io-services-cms-webapp/src/utils/subscription.ts
```diff
@@ -24,6 +29,23 @@ type ErrorResponses =
...
+  ownerId: NonEmptyString,
+  userId: NonEmptyString
+) => ownerId === userId;
+
+const extractOwnerId = (
+  fullPath?: string
+): TaskEither<IResponseErrorNotFound, NonEmptyString> =>
```
gunzip, line 39: it looks like this method is synchronous, why do we need to return TE?

apps/io-services-cms-webapp/src/utils/subscription.ts
```diff
@@ -64,8 +86,14 @@ export const serviceOwnerCheckManageTask = (
       )
     ),
     TE.chainW((serviceSubscription) =>
-      serviceSubscription.ownerId === userId
-        ? TE.right(serviceId)
-        : TE.left(ResponseErrorForbiddenNotAuthorized)
+      pipe(
+        serviceSubscription.ownerId,
+        extractOwnerId,
+        TE.chainW((ownerId) =>
```
gunzip, line 92: maybe we can use `TE.filterOrElse`?

apps/io-services-cms-webapp/src/utils/subscription.ts
giamma1295: 009e4dc

apps/io-services-cms-webapp/src/utils/subscription.ts
giamma1295: 009e4dc

apps/io-services-cms-webapp/src/utils/subscription.ts
giamma1295: We cannot use TE.filterOrElse, cause we want to return the serviceId on the right and not the ownerId, so we need the chainW

``` typescript
     pipe(
        serviceSubscription.ownerId,
        extractOwnerId,
        TE.fromEither,

        TE.filterOrElseW(
          (ownerId) => ownerId === userId, // this will return TE.right(ownerId) which is wrong
          () => ResponseErrorForbiddenNotAuthorized 
        ),
```

giuseppedipinto (approved):

PR #164, opened 2023-06-26
fix: bulk fetch on store cosmos which prevent from fetching any data

giamma1295:
Fix CosmosDb bulkFetch operation which was preventing fetch of anything

Local application run

#### Checklist:

### DIFF
--- a/packages/io-services-cms-models/src/lib/fsm/store.cosmos.ts
+++ b/packages/io-services-cms-models/src/lib/fsm/store.cosmos.ts
@@ -82,7 +82,7 @@ export const createCosmosStore = <
       TE.map((operationResponses) =>
         operationResponses.map(
           flow(
-            O.fromPredicate((res) => res.statusCode === 404),
+            O.fromPredicate((res) => res.statusCode === 200),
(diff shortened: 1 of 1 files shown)


giuseppedipinto (approved):
LGTM and thanks!

rodoviti (approved):
LGTM

PS: we could/should add some unit tests 😄

packages/io-services-cms-models/src/lib/fsm/store.cosmos.ts
```diff
@@ -82,7 +82,7 @@ export const createCosmosStore = <
       TE.map((operationResponses) =>
         operationResponses.map(
           flow(
-            O.fromPredicate((res) => res.statusCode === 404),
+            O.fromPredicate((res) => res.statusCode === 200),
```
gunzip, line 85: maybe it's worth a unit test?

giamma1295: a ticket to test store.cosmos.ts has been created

PR #173, opened 2023-06-27
chore: [IOPAE-385] add legacy filter to service lifecycle watcher

rodoviti:
Add a filter to service lifecycle watcher to avoid opening a review or publication request for services from legacy system.

#### List of Changes

- Add a filter (`item.fsm.lastTransition`) to service lifecycle watcher

#### Motivation and Context

To avoid "side effects" on synching CMS from Legacy

Unit Tests

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
+++ b/apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
@@ -71,24 +71,27 @@ export const handler =
     | ((OnSubmitActions | OnApproveActions) & RequestHistoricizationAction)
   > =>
   ({ item }) => {
-    // eslint-disable-next-line sonarjs/no-small-switch
-    switch (item.fsm.state) {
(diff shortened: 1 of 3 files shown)


giamma1295 (approved):
LGTM

apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
```diff
@@ -71,24 +71,27 @@ export const handler =
...
-        return pipe(
-          item,
-          onApproveHandler(config),
-          (x) => x,
-          (actions) => ({ ...actions, ...onAnyChangesHandler(item) }),
-          TE.right
-        );
-      default:
-        return TE.right(onAnyChangesHandler(item));
+    if (item.fsm.lastTransition !== "from Legacy") {
```
gunzip, line 74: what about having a comment that explain the reason for this check?

PR #194, opened 2023-07-05
feat: [IOPAE-430] request sync legacy handler

rodoviti:
Add OnRequestSyncLegacy Azure Function

#### List of Changes

Add OnRequestSyncLegacy Azure Function:
- definition
- entry point
- handler

#### Motivation and Context

Add this function to dequeue sync request from CSM to Legacy and to "upsert" legacy services database

Unit tests

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/main.ts
+++ b/apps/io-services-cms-webapp/src/main.ts
@@ -84,7 +84,7 @@ const fsmPublicationClient = ServicePublication.getFsmClient(
 );
 
 const servicesContainer = cosmosdbClient
-  .database(config.COSMOSDB_NAME)
+  .database(config.LEGACY_COSMOSDB_NAME)
   .container(SERVICE_COLLECTION_NAME);
 
 const serviceModel = new ServiceModel(servicesContainer);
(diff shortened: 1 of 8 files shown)


apps/io-services-cms-webapp/src/main.ts
```diff
@@ -76,6 +83,12 @@ const fsmPublicationClient = ServicePublication.getFsmClient(
   servicePublicationStore
 );
 
+const servicesContainer = cosmosdbClient
+  .database(config.COSMOSDB_NAME)
```
giuseppedipinto, line 87: it should be `config.LEGACY_COSMOSDB_NAME`, right?

giuseppedipinto (changes requested):

apps/io-services-cms-webapp/src/main.ts
rodoviti: fix with fcc7de867df813842057b7ede03843a930a72ded

giuseppedipinto (approved):
LGTM

PR #197, opened 2023-07-07
[#IOAPE-434] Added FeatureFlag for cms to legacy sync test

giamma1295:
Added a Feature Flag in order to selectively allow user to Sync Services from CMS to Legacy.
#### List of Changes
- Add `USERID_CMS_TO_LEGACY_SYNC_INCLUSION_LIST`
- Add `feature-flag-handler.ts` to manage feature flags
- Update Service History Watcher to use the new feature flag

Unit Tests

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
+++ b/apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
@@ -14,7 +14,7 @@ const wildcard = "*" as NonEmptyString;
 // Return True if the element is in the list or if the list contains the wildcard
 const isElementAllowedOnList =
   (list: ReadonlyArray<NonEmptyString>) => (element: NonEmptyString) =>
-    list.include(swildcard) || list.includes(element);
+    list.includes(wildcard) || list.includes(element);
 
 /**
  * Check if the owner of the service is enabled for CMS to Legacy Sync
--- a/apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
+++ b/apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
@@ -14,7 +14,7 @@ const wildcard = "*" as NonEmptyString;
 // Return True if the element is in the list or if the list contains the wildcard
 const isElementAllowedOnList =
   (list: ReadonlyArray<NonEmptyString>) => (element: NonEmptyString) =>
-    list.indexOf(wildcard) >= 0 || list.indexOf(element) >= 0;
+    list.include(swildcard) || list.includes(element);
 
 /**
  * Check if the owner of the service is enabled for CMS to Legacy Sync
--- a/apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
+++ b/apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
@@ -41,9 +41,9 @@ export const isUserEnabledForCmsToLegacySync = (
           `An error has occurred while retrieving service '${serviceId}'`
         )
     ),
-    TE.map((serviceSubscription) =>
+    TE.map(({ ownerId }) =>
       pipe(
-        serviceSubscription.ownerId as NonEmptyString,
+        ownerId as NonEmptyString,
         parseOwnerIdFullPath,
         isElementAllowedOnList(config.USERID_CMS_TO_LEGACY_SYNC_INCLUSION_LIST)
       )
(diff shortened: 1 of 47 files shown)


apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+  getSubscription,
+  parseOwnerIdFullPath,
+} from "../lib/clients/apim-client";
+
+const wildcard = "*" as NonEmptyString;
+
+// Return True if the element is in the list or if the list contains the wildcard
+const isElementAllowedOnList =
+  (list: ReadonlyArray<NonEmptyString>) => (element: NonEmptyString) =>
+    list.indexOf(wildcard) >= 0 || list.indexOf(element) >= 0;
```
gunzip, line 17:
```suggestion
    list.include(swildcard) || list.includes(element);
```

apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+    ),
+    TE.mapLeft(
+      (_) =>
+        new Error(
+          `An error has occurred while retrieving service '${serviceId}'`
+        )
+    ),
+    TE.map((serviceSubscription) =>
+      pipe(
+        serviceSubscription.ownerId as NonEmptyString,
```
gunzip, line 46:
```suggestion
    TE.map(({ ownerId }) =>
      pipe(
        ownerId as NonEmptyString,
```

apps/io-services-cms-webapp/src/watchers/on-service-history-change.ts
```diff
@@ -101,15 +104,28 @@ const toRequestSyncLegacyAction = (
...
-  NoAction | RequestSyncLegacyAction
-> = ({ item }) =>
-  pipe(
-    item,
-    O.fromPredicate((itm) => itm.fsm.lastTransition !== SYNC_FROM_LEGACY),
-    O.map(toRequestSyncLegacyAction),
-    O.getOrElse(() => noAction),
-    E.right
-  );
+export const handler =
```
gunzip, line 107: why not just:

`ReaderEither<{ item: ServiceHistory, config: IConfig, apimClient: ApiManagementClient}, ...>`

in this way you don't have to pass around the apimClient.

giamma1295: Cannot be dono atm, cause we need to edit processBatchOf which is used for all Handler CosmosDbTrigger

apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
giamma1295: 089d886

infra/src/key_vault.tf
```diff
@@ -33,6 +33,11 @@ data "azurerm_key_vault_secret" "serviceid_quality_check_exclusion_list" {
   key_vault_id = module.key_vault_domain.id
 }
 
+data "azurerm_key_vault_secret" "userid_cms_to_legacy_sync_inclusion_list" {
+  name         = "USERID-CMS-TO-LEGACY-SYNC-INCLUSION-LIST"
+  key_vault_id = module.key_vault_domain.id
+}
+
```
rodoviti, line 40: why a list of user id is a secret? I mean, I'm pretty sure they are not sensitive datas, so there is no need to put them in the key vault

giamma1295: bc9c699

apps/io-services-cms-webapp/src/utils/feature-flag-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param config
+ * @param apimClient
+ * @param serviceId
+ * @returns
+ */
+export const isUserEnabledForCmsToLegacySync = (
+  config: IConfig,
+  apimClient: ApiManagementClient,
+  serviceId: NonEmptyString
+) =>
```
rodoviti, line 30: add return type to "enforce" type checking

giamma1295: f5fb2fb

rodoviti (approved):

PR #222, opened 2023-07-13
fix: [IOPAE-443] OnRequestSyncCms process list items instead a single item

giamma1295:
The messages in the `request-sync-cms` contains a list of items, to be exact 2 items [serviceLifeycle, ServicePublication], so we adapted the code to handle the list of items.

Unit Test

#### Checklist:


apps/io-services-cms-webapp/src/synchronizer/request-sync-cms-handler.ts
```diff
@@ -53,34 +54,36 @@ export const handleQueueItem = (
...
-            TE.map((_) => void 0)
-          );
-        default:
-          // eslint-disable-next-line @typescript-eslint/no-unused-vars
-          const _: never = item;
-          throw new Error(`should not have executed this with ${item}`);
-      }
-    }),
+    TE.chainW(
+      RA.traverse(TE.ApplicativePar)((item) => {
```
rodoviti, line 58: **issue:** if run in parallel mode, it is possible that the FSM publication will be updated (perhaps even as the first operation) and the FSM lifecycle update will fail. This leads to an inconsistent state (e.g.: a published service, but not existing in the FSM Lifecycle), so we should run the two operations in sequence!
```suggestion
      RA.traverse(TE.ApplicativeSeq)((item) => {
```
NB: I'm not sure if the above suggestion can fix the "inconsistent state" problem!

rodoviti (changes requested):

packages/io-services-cms-models/src/queue/index.ts
```diff
@@ -47,5 +47,8 @@ export const RequestSyncCmsItem = t.union([
   ]),
 ]);
 
+export type RequestSyncCmsItemsList = t.TypeOf<typeof RequestSyncCmsItemsList>;
+export const RequestSyncCmsItemsList = t.readonlyArray(RequestSyncCmsItem);
```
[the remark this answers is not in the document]
giamma1295: 4170a0e

rodoviti (approved):

apps/io-services-cms-webapp/src/synchronizer/request-sync-cms-handler.ts
giamma1295: 74636c0

PR #231, opened 2023-07-17
fix:[BUG][#IOPAE-455]Generator doesn't generate correctly required parameter inside allOf in openApi

StefanoDoc:
With the way we wrote the openApi using an **allOf** object and the required parameter outside of it , the generator does not generate the : `id` , `status` and `last_update` for the serviceLifecycle/Publication model to be mandatory , but sets them as optional
#### List of Changes

- Moved required parameter inside the allOf object in the openApi
- update converters
- update tests

First the Old one , Second the New one

old openApi
new openApi

#### Checklist:

### DIFF
--- a/.changeset/dull-parrots-draw.md
+++ b/.changeset/dull-parrots-draw.md
@@ -1,0 +1,5 @@
+---
+"io-services-cms-webapp": patch
+---
+
+Generator doesn't generate correctly required parameter inside allOf in openApi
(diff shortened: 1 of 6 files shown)


giamma1295 (approved):
LGTM

giuseppedipinto (approved):

rodoviti (approved):

PR #244, opened 2023-07-19
fix: [IOPAE-479] unpublish service when deleting id coming from legacy sync

rodoviti:
unpublish service when deleting id coming from legacy sync

#### List of Changes

- add an unpublish item to requestSyncCms when a legacy service is deleted

Unit Test

#### Checklist:


giuseppedipinto (approved):
LGTM

PR #250, opened 2023-07-20
chore(fix): Log status response jira client

giamma1295:
Add status on checkJiraResponse when an unknown status is received
Unit Test

#### Checklist:


apps/io-services-cms-webapp/src/lib/clients/jira-client.ts
```diff
@@ -100,7 +100,9 @@ export const checkJiraResponse = (
   } else if (response.status >= 500) {
     return E.left(new Error("Jira API returns an error"));
   } else {
-    return E.left(new Error("Unknown status code response error"));
+    return E.left(
+      new Error(`Unknown status code, ${response.status}, received`)
```
giuseppedipinto:
```suggestion
      new Error(`Unknown status code ${response.status} received`)
```

giuseppedipinto (approved):

PR #275, opened 2023-07-24
chore: Add retry capabilities to Jira Legacy Client on Throttled response

giamma1295:
Add retry capabilities to Jira Legacy Client on Throttled response

#### List of Changes
- Handle `Retry-After` header on Jira response, this header contains the number of seconds which needs to be Re-Done the REST API call to the Jira API.
- Add related unit tests.

#### Motivation and Context
Manage throttled responses received from jira accordin to this [documentation page](https://developer.atlassian.com/cloud/jira/platform/rate-limiting/#rate-limit-responses).

Unit Tests

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/lib/clients/jira-legacy-client.ts
+++ b/apps/io-services-cms-webapp/src/lib/clients/jira-legacy-client.ts
@@ -5,6 +5,8 @@ import * as E from "fp-ts/lib/Either";
 import * as O from "fp-ts/lib/Option";
 import * as TE from "fp-ts/lib/TaskEither";
 import { pipe } from "fp-ts/lib/function";
+import { isNumber } from "fp-ts/lib/number";
+import * as T from "fp-ts/task";
 import * as t from "io-ts";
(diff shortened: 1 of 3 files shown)


apps/io-services-cms-webapp/src/lib/clients/jira-legacy-client.ts
```diff
@@ -125,6 +146,63 @@ export const jiraLegacyClient = (
...
+    return attempt(0);
+  };
+
+  const extractHeaders = (response: Response) => {
+    try {
+      const headers: Record<string, string> = {};
+      response.headers.forEach((value, key) => {
+        // eslint-disable-next-line functional/immutable-data
+        headers[key] = value;
+      });
```
rodoviti, line 199: Have you tried this??
```suggestion
      const headers = Object.fromEntries(response.headers.entries())
```

rodoviti (approved):
a little bit tricky, but we know we will throw it out in September 😉

PR #277, opened 2023-07-25
chore: add legacy cosmos lease for test

giamma1295:
Set `LEGACY_COSMOSDB_CONTAINER_SERVICES_LEASE` to the test one
#### Checklist:


rodoviti:
**suggestion:** @giamma1295 I "strongly" suggest to set the "test lease" collection directly from function configurations, without dirtying the code, so when we the tests are finished we can set the corretct configs without a deploy/apply 😉

PR #282, opened 2023-07-31
[#IOPAE-492]Add legacy status to jira client

StefanoDoc:
Add legacy status to jira client
#### List of Changes

add `DONE `and `completato` status 
#### Motivation and Context

we need to save on the legacy pending table the legacy ticket that have different status from the one in the new jira board , so we can use the same poller that we use for the cms jira board 
#### Checklist:


apps/io-services-cms-webapp/src/lib/clients/jira-client.ts
```diff
@@ -9,6 +9,7 @@ import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import nodeFetch from "node-fetch-commonjs";
 import { readableReport } from "@pagopa/ts-commons/lib/reporters";
 import { JiraConfig } from "../../config";
+// import { JiraLegacyIssueStatus } from "./jira-legacy-client";
```
rodoviti, line 12:
```suggestion
```

rodoviti (changes requested):

rodoviti (approved):

giuseppedipinto (approved):

PR #286, opened 2023-08-01
[#IOPAE-510] Removed Jira call and update status mapping on LegacyServiceWatcher

giamma1295:
Removed Jira call and update status mapping on LegacyServiceWatcher
#### List of Changes
- Removed `Jira REST API call` on `LegacyServiceWatcher`
- Add Fetch `previous service Version` to correctly intercept `unpublish actions` on legacy.
- Update `status mapping` on `LegacyServiceWatcher`
- Update `LegacyServiceWatcher Unit Test` according to the changes

#### Motivation and Context
When a service is submitted for a review or a review is rejected, on the legacy service management application, the service Itself will not change, so no Change Feed notification will be intercepted by this watcher, this means that we need to intercept the creation of new Jira Ticket and Jira ticket status changes on dedicated Functions to fully sync those kind of changes on legacy and sync in io-services-cms.

Unit Tests

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
+++ b/apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
@@ -103,10 +103,8 @@ const legacyToCms = (item: LegacyService, legacyServiceModel: ServiceModel) =>
     ),
     TE.map(
       flow(
-        O.fold(
-          () => getLegacyToCmsStatus(item),
-          (wasPublished) => getLegacyToCmsStatus(item, wasPublished)
-        )
+        O.getOrElse(() => false),
+        (wasPublished) => getLegacyToCmsStatus(item, wasPublished)
       )
     ),
     TE.map((statusList) =>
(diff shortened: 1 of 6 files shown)


apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
```diff
@@ -167,11 +114,59 @@ const legacyToCms = (
     )
   );
 
+const serviceWasPublished = (
+  item: LegacyService,
+  legacyServiceModel: ServiceModel
+): TE.TaskEither<Error, O.Option<boolean>> =>
```
rodoviti, line 120: question: why are you returning an Optional? Can't we return a boolean directly?

giamma1295: Changed in d34deed, now we map the method result into an Option when needed

giuseppedipinto (approved):

rodoviti (approved):

PR #289, opened 2023-08-02
[#IOPAE-493] Service review legacy checker

giamma1295:
#### List of Changes
Add New `ServiceReviewLegacyChecker`, Azure Function that will sync any Jira Legacy status change to the imported legacy services.

- ADD ServiceReviewLegacyChecker AF `function.json`
- ADD AzureWebJobs.ServiceReviewLegacyChecker.Disabled in `webapp.tf`
- ADD `ServiceReviewLegacyChecker` Azure Function
- ADD Unit Test for `ReviewLegacyCheckerHandler`
#### Motivation and Context

When a Service is Submitted for an approval on Legacy Service a Jira Ticket is created, the workflow on those tickets needs to be tracked in order to correctly intercept all Legacy service status change and sync them in `io-services-cms`

Unit Test

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/main.ts
+++ b/apps/io-services-cms-webapp/src/main.ts
@@ -17,6 +17,7 @@ import * as O from "fp-ts/Option";
 import * as RA from "fp-ts/ReadonlyArray";
 import * as RR from "fp-ts/ReadonlyRecord";
 import { pipe } from "fp-ts/lib/function";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import { getConfigOrThrow } from "./config";
 import { createRequestHistoricizationHandler } from "./historicizer/request-historicization-handler";
 import {
@@ -47,6 +48,7 @@ import { handler as onServicePublicationChangeHandler } from "./watchers/on-serv
 import { createWebServer } from "./webservice";
 
 import { initTelemetryClient } from "./utils/applicationinsight";
+import { createReviewLegacyCheckerHandler } from "./reviewer/review-legacy-checker-handler";
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-unused-vars
 const BASE_PATH = require("../host.json").extensions.http.routePrefix;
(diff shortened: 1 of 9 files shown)


apps/io-services-cms-webapp/src/main.ts
```diff
@@ -140,6 +142,18 @@ export const serviceReviewCheckerEntryPoint = createReviewCheckerHandler(
   fsmLifecycleClient
 );
 
+export const serviceReviewLegacyCheckerEntryPoint =
+  createReviewLegacyCheckerHandler(
+    getDao({
+      ...config,
+      REVIEWER_DB_TABLE: `${config.REVIEWER_DB_TABLE}-legacy` as NonEmptyString,
```
rodoviti, line 149: **issue (blocking):** replace _dash_ with _underscore_
```suggestion
      REVIEWER_DB_TABLE: `${config.REVIEWER_DB_TABLE}_legacy` as NonEmptyString,
```

apps/io-services-cms-webapp/src/reviewer/review-checker-handler.ts
```diff
@@ -106,44 +106,40 @@ export const buildIssueItemPairs =
  */
 export const updateReview =
   (dao: ServiceReviewDao, fsmLifecycleClient: ServiceLifecycle.FsmClient) =>
-  (data: TE.TaskEither<Error, IssueItemPair[]>): TE.TaskEither<Error, void> =>
+  (issueItemPairs: IssueItemPair[]): TE.TaskEither<Error, void> =>
     pipe(
```
rodoviti, line 110: **suggestion (non-blocking):** use _flow_.
You could use _flow_ instead of _pipe(issueItemPairs_
```suggestion
    flow(
```

apps/io-services-cms-webapp/src/reviewer/review-checker-handler.ts
```diff
@@ -106,44 +106,40 @@ export const buildIssueItemPairs =
...
-                ...item,
-                status: issue.fields.status.name,
-              }),
-              TE.mapLeft((err) => {
-                // eslint-disable-next-line no-console
-                console.error(err.message); // FIXME: is it correct to log via console?
-                return E.toError(err);
-              })
-            )
+      issueItemPairs,
```
rodoviti, line 111: **suggestion (non-blocking):** use _flow_.
You could use _flow_ instead of _pipe(issueItemPairs_
```suggestion
```

rodoviti (changes requested):

apps/io-services-cms-webapp/src/reviewer/review-checker-handler.ts
giamma1295: In this case pipe cannot be replaced by flow as is the start of an arrow function

apps/io-services-cms-webapp/src/reviewer/review-checker-handler.ts
giamma1295: In this case pipe cannot be replaced by flow as is the start of an arrow function

rodoviti (approved):

giuseppedipinto (approved):

PR #306, opened 2023-08-23
fix: Fetch previous version recursive

giamma1295:
On legacy service versions are not consecutive, so we have to recursively find it
#### Checklist:

### DIFF
--- a/.changeset/eight-items-drive.md
+++ b/.changeset/eight-items-drive.md
@@ -1,0 +1,5 @@
+---
+"io-services-cms-webapp": patch
+---
+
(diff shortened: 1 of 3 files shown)


rodoviti (approved):

.changeset/eight-items-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"io-services-cms-webapp": patch
+---
+
+Fetch previous legacy service recursive
```
gunzip, line 5: recursive what? :-)

PR #310, opened 2023-08-23
[IOPAE-506] Add BackOffice openapi.yaml

giuseppedipinto:
This is the IO BackOffice Backend API specification based on the OpenAPI 3.0.

#### List of Changes
- Added **openapi.yaml** specification
- Updated **package.json** with `generate` script and `@pagopa/openapi-codegen-ts` devDependency

#### Checklist:


apps/backoffice/openapi.yaml
```diff
@@ -0,0 +1,817 @@
...
+    get:
+      tags:
+        - auth
+      summary: Resolve selfcare identity
+      description: Resolve selfcare identity
+      operationId: resolveSelfCareIdentity
+      security: [] # public endpoint
+      parameters:
+        - name: identity-token
+          in: header
```
rodoviti, line 68: **nitpick:** I prefer to pass it as a request body instead of a header param (to try to stay more "closely" to the oauth protocol)

giuseppedipinto: a42e8f6

rodoviti (approved):

PR #322, opened 2023-08-25
chore: default value for departmentName before sync to legacy

giamma1295:
Field `organization.department_name` on io-services-cms is optional, but the corresponding field, `departmentName`, on the old system is mandatory, so we introduce a placeholder value before sync a service from io-services-cms -> legacy when undefined.
Unit Test

#### Checklist:


apps/backoffice/package.json
```diff
@@ -37,4 +37,4 @@
     "typescript": "5.1.6",
     "vitest": "^0.34.1"
   }
-}
\ No newline at end of file
+}
```
rodoviti, line 40: **question (non-blocking):** is this a "wanted" change?

rodoviti (approved):

apps/backoffice/package.json
giamma1295: The linter did this modification

.changeset/wise-onions-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"io-services-cms-webapp": patch
+---
+
+Placeholder value for field department name before syncing a service to legacy who lack's of
```
gunzip, line 5:
```suggestion
Add a placeholder value before syncing a service to legacy one that lacks the field "department name"
```

PR #333, opened 2023-09-01
chore: Export Selfcare Identity payload from inline to an external Schemas

giamma1295:
`openapi-codegen-ts` module is ignoring inline schemas, so in order to properly generate them we need to export them to an external Schema.

#### List of Changes
- moved `/auth` requestBody schema from inline to an external one `SelfCareIdentity`

#### Checklist:

### DIFF
--- a/.changeset/dull-scissors-learn.md
+++ b/.changeset/dull-scissors-learn.md
@@ -1,0 +1,5 @@
+---
+"@io-services-cms/backoffice": patch
+---
+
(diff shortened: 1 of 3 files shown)


giuseppedipinto (approved):
LGTM

rodoviti (approved):
LGTM

PS: we could leave SelfcareIdentity schema definition inside openapi.yaml instead creating a new definition file with only that

PR #346, opened 2023-09-05
[IOPAE-544] Add useFetch react custom hook

giuseppedipinto:
**Custom useFetch hook for client fetch API**
This hook use `@pagopa/openapi-codegen-ts` autogenerated client, therefore makes available all and only APIs specified in the project root opeanapi.yaml _(by operationId)_.

_Parameters:_
- `operationId` openapi operationId
- `requestParams` request parameters _(as specified in openapi)_
- `responseCodec` io-ts expected response codec

_Returns:_
- `data` fetch result of _responseCodec_ type
- `error` in case of request error, decode failure or exceptions
- `loading` as boolean value for request pending state

**Usage example:**
```
const { data, error, loading } = useFetch(
  "regenerateServiceKey",
  {
    serviceId: "aServiceId",
    keyType: SubscriptionKeyTypeEnum.primary
  },
  SubscriptionKeys
);
```
_Error Typing:_
```
export type UseFetchError = {
  kind: "httpError" | "validationError" | "exceptionError";
  status?: number; // is the http response status code
  message: string;
};
```

**Important Notes**
The client is already set up to send the **bearer token** _(currently there is a placeholder here)_ and in case of **401 Unauthorized** it manages the redirect to a dedicated route _(currently there is a placeholder here)_.

#### Checklist:


apps/backoffice/src/hooks/use-fetch.ts
```diff
@@ -0,0 +1,173 @@
...
+      status
+    });
+
+  const fetchData = async (requestParams: any) => {
+    try {
+      const result = await clientWithBearerToken[operationId](requestParams);
+
+      if (result._tag === "Right") {
+        // Check 401 Unauthorized
+        if (result.right.status === 401) {
```
gunzip, line 124: I suggest to assign a variable with a ameaningful name to `result.right`

apps/backoffice/src/hooks/use-fetch.ts
```diff
@@ -0,0 +1,173 @@
...
+        /** Set a response class type based on http response status code */
+        const httpResponseClassType = manageHttpResponseStatusCode(
+          result.right.status
+        );
+
+        switch (httpResponseClassType) {
+          case "successful":
+            // 2XX successful response status code
+            const decodedResult = responseCodec.decode(result.right.value);
+            decodedResult._tag === "Right"
```
gunzip, line 138: fold / bimap?

apps/backoffice/src/hooks/use-fetch.ts
giuseppedipinto: 8376da1

apps/backoffice/src/hooks/use-fetch.ts
giuseppedipinto: 7240f27

apps/backoffice/src/hooks/use-fetch.ts
```diff
@@ -0,0 +1,183 @@
...
+            );
+            break;
+          case "clientError":
+          case "serverError":
+            setUseFetchError(
+              "httpError",
+              httpResponseClassType,
+              response.status
+            );
+          default:
```
rodoviti, line 162: **suggestion (non-blocking):** remove "useless" default.

If a default block does not have an implementation, simply omit it.
```suggestion
```

rodoviti (approved):
LGTM

apps/backoffice/src/hooks/use-fetch.ts
giuseppedipinto: 925cfa7

apps/backoffice/src/hooks/use-fetch.ts
```diff
@@ -0,0 +1,183 @@
...
+      kind,
+      message,
+      status
+    });
+
+  const fetchData = async (requestParams: any) => {
+    try {
+      const result = await clientWithBearerToken[operationId](requestParams);
+
+      if (result._tag === "Right") {
```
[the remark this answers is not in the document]
giuseppedipinto: 07fbc1c
_(as we have seen and as the autogenerated client response is defined, it is easier to check the left branch first)_

PR #354, opened 2023-09-12
[#IOPAE-579]Update jira ticket description

StefanoDoc:
Change jira ticket description for a service review based on Operations feedbacks
#### List of Changes

- add head "Dati Servizio"
- moved service scope
- new label for service name , description and scope
- set account type to api v2 until Selfcare's integration with the new api

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/utils/jira-proxy.ts
+++ b/apps/io-services-cms-webapp/src/utils/jira-proxy.ts
@@ -13,7 +13,7 @@ import {
 } from "../lib/clients/jira-client";
 
 const formatOptionalStringValue = (value?: string) =>
-  value || `{color:#000000}*[ DATO MANCANTE ]*{color}`;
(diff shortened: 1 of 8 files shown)


apps/io-services-cms-webapp/src/utils/jira-proxy.ts
```diff
@@ -13,7 +13,7 @@ import {
 } from "../lib/clients/jira-client";
 
 const formatOptionalStringValue = (value?: string) =>
-  value || `{color:#FF991F}*[ DATO MANCANTE ]*{color}`;
+  value || `{color:#000000}*[ DATO MANCANTE ]*{color}`;
```
giuseppedipinto, line 16: If we don't need to have a specific color, we could remove it:

```suggestion
  value || `*[ DATO MANCANTE ]*`;
```

PR #358, opened 2023-09-17
[IOPAE-595] Update next-auth with session token data

giuseppedipinto:
Update frontend auth session with JWT session token target data.

#### List of Changes
- Updated `[...nextauth]/route.ts` handler with jwt session token data _(before there was only mock data)_
- Added `SessionTokenPayload ` io-ts type
- Updated `useFetch()` hook to use Bearer token data taken from jwt session token
- Updated `BackOfficeUser` and `BackOfficeUserOrganization` augmented types in `next-auth.d.ts`
- Added some translations for `operator` and `admin` roles
- Other minor corrections

#### Checklist:


apps/backoffice/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -1,5 +1,32 @@
...
+      .map(c => "%" + ("00" + c.charCodeAt(0).toString(16)).slice(-2))
+      .join("")
+  );
+
+  return pipe(codec.decode(JSON.parse(jsonPayload)));
+};
+
+/** Utility to check if a string is null, undefined or empty */
+const isNullUndefinedOrEmpty = (value: string) =>
+  value === null || value === undefined || value === "";
```
Comment (giamma1295) [inline apps/backoffice/src/app/api/auth/[...nextauth]/route.ts:29]: (non-blocking):
a simplified  version:
const isNullUndefinedOrEmpty = (value: string | null | undefined) => 
  !value || value.trim().length === 0;

giamma1295 (approved):
LGTM

apps/backoffice/src/app/api/auth/[...nextauth]/route.ts
Comment (giuseppedipinto) [reply apps/backoffice/src/app/api/auth/[...nextauth]/route.ts]: a5d2e1a

PR #361, opened 2023-09-19
[IOPAE-597] Add AccessControl component and setup unit tests frontend

giuseppedipinto:
Added **AccessControl** wrapper for content rendering based on user permissions match.
Taken this opportunity to configure vitest for **frontend unit tests**.
Added AccessControl component unit tests.

**Note:** _Follow commits history to better understand the 3 steps described above._

#### Motivation and Context
- A way to render components based on logged user permissions.
- The need to have unit tests also for frontend code.

Locale Dev, Unit tests.

#### Checklist:


apps/backoffice/src/components/access-control/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  requiredPermissions,
+  renderNoAccess,
+  children
+}: AccessControlProps) => {
+  const { data: session } = useSession();
+
+  const isArraySubset = (subset: string[], superset: string[]) =>
+    subset.every(item => superset.includes(item));
+
+  const canBeShown = () =>
```
gunzip, line 30: maybe `hasRequiredPermission` would be clearer?

giuseppedipinto: it could.
9b81b9b

StefanoDoc (approved):

PR #375, opened 2023-09-26
fix:[#IOPAE-638] Handle Jira Issue status name case change

giamma1295:
Handle changed jira issue status names
Jira api seems to be inconsistent, the field `issue.fields.status.name` sometimes is returned in UPPERCASE, sometime in PascalCase, so we are removing the field validator(which was an union of io-ts literals, which are case sensitive), and normalize it when is needed.
#### List of Changes
- Remove Type JiraIssueStatus`type literal.
- Add `decodeJiraIssueStatus` in `review-checker-handler.ts` to normalize the value before saving it on DB.

Unit test

#### Checklist:

### DIFF
--- a/.changeset/famous-ducks-cry.md
+++ b/.changeset/famous-ducks-cry.md
@@ -2,4 +2,4 @@
 "io-services-cms-webapp": patch
 ---
 
-Handle changed jira issue status names
+Handle Jira Issue status name case change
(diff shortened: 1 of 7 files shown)


rodoviti (approved):
LGTM

StefanoDoc (approved):

giuseppedipinto (approved):

PR #401, opened 2023-10-05
fix : [#IOPAE-659] bug on delete api with never approved service

StefanoDoc:
we discovered a bug on the delete api where if you called the delete api on a service that has never been approved so is not in the ServicePublication when after the delete the unpublish gets called we would throw an error because we could not find the sevice but this is a positive case for us 
unit tests

#### Checklist:


apps/io-services-cms-webapp/src/publicator/__tests__/request-publication-handler.test.ts
```diff
@@ -129,9 +133,32 @@ describe("Service Publication Handler", () => {
...
+      id: aService.id,
+      kind: "RequestUnpublicationItem",
+    } as unknown as Json;
+    const context = createContext();
+
+    class FSMError extends Error {
+      public kind = "FsmItemNotFoundError";
+    }
+    const mockFsmPublicationClient = {
+      unpublish: vi.fn(() => TE.right(new FSMError())),
```
giamma1295, line 153: Shouldn't it be TE.left in case of an error when applying a transition?

StefanoDoc: i get i serialized error if i use TE.left and i don't know why...

giamma1295: in order to solve the issue you can:
- Remove the FSMError class
- Update mockFsmPublicationClient in the following way:
``` ts
    const mockFsmPublicationClient = {
      unpublish: vi.fn(() => TE.left(new FsmItemNotFoundError(aService.id))),
    } as unknown as ServicePublication.FsmClient;
```

giamma1295 (approved):

PR #407, opened 2023-10-09
[#IOPAE-636] Implements GET /keys/manage/cidrs endpoint

giamma1295:
#### Checklist:


apps/backoffice/src/app/api/lib/authorized-cidrs.ts
```diff
@@ -0,0 +1,47 @@
...
+        )
+      )
+    ),
+    TE.map(response =>
+      NextResponse.json({
+        cidrs: response.cidrs
+      })
+    ),
+    TE.toUnion
+  );
```
rodoviti, line 47: **suggestion (non-blocking):** move the NextResponse to the caller (request handler).

I suggest to remove from this lib/util the responsibility for returning the HTTP response, but instead return a structured error object to the caller (the NextJS request handler), so that it is able to create the correct HTTP response

giamma1295: To make the controllers easily testable, I expressly chose in the design phase of these APIs not to write any logic within the methods of the various route.ts of next.js.
This choice was made because of the framework's lack of dependency injection, which makes it difficult to write unit tests.

The controller:
- obtains the dependency instances necessary for execution
- calls the corresponding a method that contains the controller logic by injecting the dependencies and request data.

The method in question executes:
- a call to a business method (whose dependencies are injected), which fetches the data from cosmos
- based on the result from cosmos composes the response, which is immediately returned in the controller

PR #409, opened 2023-10-10
[#IOPAE-675]Add delay step for backoffice action

StefanoDoc:
currently there is no delay between the deploy of the BO and the healtcheck and this causes the healthcheck to return a 404 because the BO is not up immediately after , on top of that github action do not provide a way to retry a command so the easy solution for now is to add a delay between the 2
#### List of Changes

- add a check to see if BO doesn't replie with a 200
- add delay 
- curl re-execution 

#### Checklist:


rodoviti (commented):
**question (non-blocking):** honestly, i'm not sure this "workaround" would fix the problem, because if the app loading is lazy, then it will start to load only when receiving the first "call".

.github/workflows/azure-deploy.yaml
```diff
@@ -104,6 +104,12 @@ jobs:
         run: |
           # fails if response status is not 2xx
           curl -f 'https://${{ env.AZURE_BACKOFFICE_NAME }}-staging.azurewebsites.net/${{ env.BACKOFFICE_HEALTHCHECK_PATH }}'
+          resp_status="$(curl --silent --head https://selfcare.io.pagopa.it/api/info | awk '/^HTTP/{print $2}')"
```
rodoviti, line 107: **chore:** fix first curl execution

PS: suggestion not verified, checks it before accept! 😉 
```suggestion
          resp_status="$(curl --silent --head https://${{ env.AZURE_BACKOFFICE_NAME }}-staging.azurewebsites.net/${{ env.BACKOFFICE_HEALTHCHECK_PATH }} | awk '/^HTTP/{print $2}')"
```

rodoviti (changes requested):

rodoviti (approved):
LGTM

PR #411, opened 2023-10-12
fix : add final error status check for BO deploy action

StefanoDoc:
after adding retry step to the BO deploy github action we add a final check to interrupt the flow in case after all the retry we still have an error
#### Checklist:

### DIFF
--- a/.github/workflows/azure-deploy.yaml
+++ b/.github/workflows/azure-deploy.yaml
@@ -113,7 +113,7 @@ jobs:
           done
           if [[ $resp_status !== 200 ]] 
           then
-            echo Not Healty HTTP status $resp_status after $retryCount retries
+            echo "Not Healty (response status $resp_status) after $retryCount retries"
             exit 1
           fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/azure-deploy.yaml
```diff
@@ -111,6 +111,11 @@ jobs:
             echo resp_status="$(curl --silent --head https://${{ env.AZURE_BACKOFFICE_NAME }}-staging.azurewebsites.net/${{ env.BACKOFFICE_HEALTHCHECK_PATH }} | awk '/^HTTP/{print $2}')"
             retryCounter=$(( $retryCounter + 1 ))
           done
+          if [[ $resp_status !== 200 ]] 
+          then
+            echo Not Healty HTTP status $resp_status after $retryCount retries
```
rodoviti, line 116: **chore:** fix message
```suggestion
            echo "Not Healty (response status $resp_status) after $retryCount retries"
```

rodoviti (changes requested):

rodoviti (approved):

PR #422, opened 2023-10-17
[#IOPAE-684] Narrow criteria sync cms to legacy

StefanoDoc:
You can find information on why we need this pr [here](https://pagopa.atlassian.net/browse/IOPAE-681) , basically we found a bug in the synchronization mechanism so we need to narrow the criteria of it

#### List of Changes
change the sync towards the Service db so that now we sync only if 

- entry is ServicePublication fsm type or is ServiceLifecycle and approved
- entry is ServiceLifecycle and it doesn't exist a ServicePublication entry of it

unit tests

#### Checklist:


giamma1295:
Not Necessary cause another strategy was choosen

PR #430, opened 2023-10-19
feat: [IOPAE-676] apim user creation

rodoviti:
If user (i.e. the institution) not exists, create a new Apim user (i.e. the institution) and add to it the defaults groups

#### List of Changes

- add user creation if not exists in token exchange process
- after user creation, add default groups to it
- add mock

#### Motivation and Context

In case of first user access (selfcare), it must be created to apim with default permission

#### Checklist:


apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts
```diff
@@ -65,21 +68,90 @@ const retrieveApimUser = (config: Configuration) => (
...
-        apimClient,
-        config.AZURE_APIM_RESOURCE_GROUP,
-        config.AZURE_APIM
-      ),
-    apimService =>
-      apimService.getUserByEmail(
-        `org.${identityTokenPayload.organization.id}@selfcare.io.pagopa.it` as EmailString,
-        true
+      pipe(
+        ApimUtils.getApimService(
```
Comment (giamma1295) [inline apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts:72]: Non Blocking, you can use getApimService from`@lib/be/apim-service` to get an instance of ApimService

apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts
```diff
@@ -65,21 +68,90 @@ const retrieveApimUser = (config: Configuration) => (
...
+            )
+        ),
+        TE.chain(
+          flow(
+            O.fold(
+              () =>
+                pipe(
+                  TE.tryCatch(
+                    () =>
+                      apimClient.user.createOrUpdate(
```
Comment (giamma1295) [inline apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts:97]: Non Blocking, this method can be added to external-service/apim-client in order to have all interaction with Apim in a single place

giamma1295 (approved):
Some Non-Blocking suggestion has been disclosed, in general seems OK

apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts
```diff
@@ -65,21 +68,90 @@ const retrieveApimUser = (config: Configuration) => (
...
+                          note: identityTokenPayload.organization.fiscal_code,
+                          appType: "backofficeIO"
+                        }
+                      ),
+                    E.toError
+                  ),
+                  TE.chain(apimUser =>
+                    TE.tryCatch(
+                      () =>
+                        Promise.all(
```
Comment (gunzip) [inline apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts:116]: this makes parallel calls, ensure that APi management can handle these without throttllng. more idiomatic: consider chain Taskeithers obtained using `TE.sequenceSeqArray`

apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts
Comment (rodoviti) [reply apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts:72]: @giamma1295 I know, but I would like to introduce no more changes than are strictly necessary (related to this task). I will handle this refactor in a dedicated PR 😉

apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts
Comment (rodoviti) [reply apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts:97]: @giamma1295 I know, but I would like to introduce no more changes than are strictly necessary (related to this task). I will handle this refactor in a dedicated PR 😉

apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts
Comment (rodoviti) [reply apps/backoffice/src/app/api/auth/[...nextauth]/auth.ts]: implemented with 7619510fce0ca50906382463aba09df4bfdda7e2

PR #448, opened 2023-10-26
[IOPAE-723] B4F Healthchecks

giamma1295:
Implementing HealthChecks for the backoffice B4F.

#### List of Changes
- ADD new Error `HealthCheckError` in order to handle Health checks
- ADD `getApimHealth` a method for checking the operating status with `APIM`
- ADD `getIoServicesCmsHealth` a method for checking the operating status with `io-services-cms` 
- ADD `getLegacyCosmosHealth` a method for checking the operating status with `io-p-cosmos-api` 
- ADD `getSelfcareHealth` a method for checking the operating status with `selfcare`
- ADD `healthcheck.ts` which is responsible for health checks handling.
- ADD healthchecks information in `GET /api/info` Route Handler

Example Info Response:

1. No Healthcheck failures
```json
{
    "name": "io-services-cms-backoffice",
    "version": "0.11.0",
    "health": {
        "status": "ok"
    }
}
```
2. One or more Healthchecks fails
```json
{
    "name": "io-services-cms-backoffice",
    "version": "0.11.0",
    "health": {
        "status": "fail",
        "failures": [
            {
                "service": "legacy-cosmos-db",
                "errorMessage": "error message"
            },
            {
                "service": "cms-client",
                "errorMessage": "error message"
            }
        ]
    }
}
```

Unit Tests

#### Checklist:

### DIFF
--- a/.changeset/violet-cobras-turn.md
+++ b/.changeset/violet-cobras-turn.md
@@ -1,0 +1,5 @@
+---
+"io-services-cms-backoffice": patch
+---
+
+ADD HealthChecks to backoffice B4F
--- a/apps/backoffice/openapi.yaml
+++ b/apps/backoffice/openapi.yaml
@@ -42,20 +42,13 @@ paths:
           content:
             application/json:
               schema:
-                type: object
-                properties:
-                  version:
-                    type: string
-                    description: the application version
-                  name:
-                    type: string
-                    description: the application name
+                $ref: '#/components/schemas/Info'
         '500':
           description: The fails to startup due to misconfiguration or unreachable dependencies
           content:
             application/json:
               schema:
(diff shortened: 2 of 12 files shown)


giuseppedipinto (approved):

rodoviti (commented):
can we create a new health-check endpoint (i.e. `/api/health`) witch return a "tiny" response like the following:
`{ status: "ok" }`
or
`{ status: "ko", failures: ["apim"] }`
After which, of course, we should change the health-check path in both the Backoffice App Service and Application Gateway.

This way we can save on cost and bandwidth 😉 

PS: we still leave all the helth-check details in the info endpoints so we can call it if we want to know unhealty details

PR #456, opened 2023-10-30
[#IOPAE-749] ADD GET Service List API in Backoffice B4F

giamma1295:
**ADD GET Service List API in Backoffice B4F**, this new api provides the user service list paginated.
#### List of Changes
- Moved `@/lib/be/cms` -> `@/lib/be/services`
- Add new Route Handler `/services/list`
- Implements `GET` method on Route handler `/services/list`
- Add `@/lib/be/services/business.retrieveServiceList` function, this function serve the new route handler mentioned above.
- Add unit test for `@/lib/be/services/business.retrieveServiceList` function
- ADD `AZURE_CREDENTIALS_SCOPE_URL` config in `backoffice_webapp.tf`
- ADD `AZURE_APIM_SUBSCRIPTIONS_API_BASE_URL` in `backoffice_webapp.tf`

Unit Test and Local application run

#### Checklist:


apps/backoffice/src/lib/be/apim-service.ts
```diff
@@ -42,6 +49,73 @@ export const getApimService: () => ApimUtils.ApimService = cache(() => {
...
+  if (!accessToken) {
+    throw new Error("invalid or null accessToken");
+  }
+
+  return axios.create({
+    baseURL: apimConfig.AZURE_APIM_SUBSCRIPTIONS_API_BASE_URL,
+    timeout: 5000,
+    headers: { Authorization: `Bearer ${accessToken}` }
+  });
+});
```
rodoviti, line 78: **question:** this rest client instance is cached, but what if the [AccesToken](https://learn.microsoft.com/en-us/javascript/api/@azure/identity/accesstoken) retrieved by [getToken](https://learn.microsoft.com/en-us/javascript/api/@azure/identity/clientsecretcredential?view=azure-node-latest#@azure-identity-clientsecretcredential-gettoken) during Axios instance creation expired? (currently, I think the SDK handles this automatically)

apps/backoffice/.env.development
```diff
@@ -32,6 +32,10 @@ AZURE_APIM=AZURE_APIM
 APIM_USER_GROUPS=apimessagewrite,apiinforead,apimessageread,apilimitedprofileread,apiservicewrite
 API_APIM_MOCKING=true
 
+AZURE_CREDENTIALS_SCOPE_URL=https://management.azure.com/.default
+AZURE_APIM_SUBSCRIPTIONS_API_BASE_URL=https://management.azure.com/subscriptions/
```
rodoviti, line 36: chore: set this env vars also from terraform

rodoviti (changes requested):

apps/backoffice/src/lib/be/apim-service.ts
giamma1295: cache is now removed, will follow a PR with a strategy to solve this issue properly

apps/backoffice/.env.development
giamma1295: Added in 765804e5972245e00257df1fde092b26255bf3ca

rodoviti (approved):
LGTM

PR #460, opened 2023-10-31
chore: [IOPAE-715] add institution fields to createService and updateService

rodoviti:
Make the setting of organization fields transparent to the user when creating and updating the service.

#### List of Changes

- Add `organization.name` and `organization.fiscal_code` as user session fields
- Remove `organization` property from create and update service request body (B4F API)
- Add `organization` fields (picked from user session) from create and update service request body before forward request to services-cms

#### Motivation and Context

Improve user experience and system robustness

#### Checklist:

### DIFF
--- a/.changeset/cool-wombats-reply.md
+++ b/.changeset/cool-wombats-reply.md
@@ -1,0 +1,5 @@
+---
+"io-services-cms-backoffice": minor
+---
+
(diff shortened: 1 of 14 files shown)


giuseppedipinto (approved):
LGTM

PR #481, opened 2023-11-07
[#IOPAE-776] Handle 404 apim rest response

giamma1295:
Handle Apim 404 Response on getServiceList, we will return an empty list.
Unit tests

#### Checklist:


apps/backoffice/src/lib/be/services/business.ts
```diff
@@ -19,6 +19,8 @@ import {
   retrievePublicationServices
 } from "./cosmos";
 import { reducePublicationServicesList, toServiceListItem } from "./utils";
+import { getServiceList } from "./apim";
+import { use } from "i18next";
```
giuseppedipinto, line 23: Is it a typo?

giuseppedipinto (approved):

apps/backoffice/src/lib/be/services/business.ts
giamma1295: yes, removed, thank's!

PR #488, opened 2023-11-08
[#IOPAE-759] change jira contract field on create to Null

StefanoDoc:
Change requested by L2 , the contract field when creating a ticket needs to be `Null` and not `Assente`
#### Checklist:


giamma1295 (approved):
LGTM

PR #491, opened 2023-11-08
[IOPAE-752] Frontend services migration: import from Developer Portal

giuseppedipinto:

https://github.com/pagopa/io-services-cms/assets/118166285/667230a5-d5d5-4e5b-ae0c-7966f1351008

#### Checklist:


apps/backoffice/src/components/services/subscriptions-migration/migration-latest.tsx
```diff
@@ -0,0 +1,113 @@
...
+      ? { label: "failed", color: "error" }
+      : // at least one still processing means the overall process still processing
+      status.processing > 0
+      ? { label: "doing", color: "warning" }
+      : // for every other case, we consider it done
+        { label: "done", color: "success" };
+  };
+
+  useEffect(() => {
+    migrationItems?.items?.length === 0
```
gunzip, line 46:
```suggestion
    setIsEmptyState(migrationItems?.items?.length === 0)
```

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,196 @@
+import { ButtonWithLoader } from "@/components/buttons";
```
gunzip, line 1: can you add a top level comment about what this module does?

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,196 @@
...
+        return { data: JSON.stringify(error), status: -1 };
+      }
+    }
+  };
+
+  // direct axios fetch
+  const postDelegateMigrationOwnershipClaims = async (delegateId: string) => {
+    try {
+      const { status } = await axios.post<void>(
+        getMigrationOwnershipClaimsUrl(delegateId)
```
gunzip, line 80: why the name is post but the call here is get?

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  // for each delegate fetch own migration status
+  useEffect(() => {
+    fetchDelegateStatusPairs();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [migrationDelegatesData]);
+
+  // When import modal is open, fetch delegate list
+  useEffect(() => {
+    if (isImportModalOpen) {
```
gunzip, line 145: why don't you put this code into `handleOpenImportModal` instead than `useEffect`?

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: It's composed url get (It has nothing to do with post as an http method)

apps/backoffice/src/components/services/subscriptions-migration/migration-latest.tsx
giuseppedipinto: Sure! d429f64

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: because in this case I chose to use an `effect-based` pattern _(precisely waiting for the modal to change state)_ which can also be triggered by other components.
In this way I rely, with certainty, on the open state of the modal.

apps/backoffice/src/components/services/subscriptions-migration/migration-latest.tsx
```diff
@@ -0,0 +1,111 @@
...
+      ? { label: "doing", color: "warning" }
+      : // for every other case, we consider it done
+        { label: "done", color: "success" };
+  };
+
+  useEffect(() => {
+    setIsEmptyState(migrationItems?.items?.length === 0);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [migrationItems]);
+
```
gunzip, line 49: Wouldn't this work in the same way? (instead of using state)

```suggestion
const isEmptyState = (migrationItems?.items?.length === 0)
```

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
gunzip: I'm not sure I get this answer. `isModelOpen` seems to be set only by `onOpenChange`. What's the difference between calling this code directly inside `onOpenChange` handler?

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: Sure! 08c6376

apps/backoffice/src/components/services/subscriptions-migration/migration-latest.tsx
giuseppedipinto: it's not a good practice to do a direct assignment that way, especially if the value changes dynamically during renders

gunzip: on the contrary, the best practice is to avoid `useEffect` whenever is it possible, expecially with derived properties. 

https://stackoverflow.com/questions/68166155/consequences-of-using-computed-variables-vs-usestate-useeffect

> The only reason to use hooks or other state tracking functionality is if you plan on changing the state within the component itself.

gunzip: see also https://prateeksurana.me/blog/why-you-should-avoid-using-state-for-computed-properties/

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
gunzip: (please don't resolve comments without code changes by yourself, otherwise I have to open the accordion for every other comment : -)

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,205 @@
...
+      );
+      return { status };
+    } catch (error) {
+      return { status: -1 };
+    }
+  };
+
+  const fetchDelegateStatusPairs = () => {
+    const pairs: Array<DelegateStatusPair> = [];
+    migrationDelegatesData?.delegates?.forEach(async delegate => {
```
gunzip, line 99: this makes parallel calls, is this what we want here?

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,205 @@
...
+      const result = await getDelegateMigrationStatus(delegate.sourceId);
+      if (result.status === 200) {
+        pairs.push({
+          delegate,
+          data: result.data
+        });
+      }
+      setMigrationStatusList([...pairs]);
```
gunzip, line 107: I'm not sure I get this, do we overwrite the status for each iteration? what about setting it once in the end (outside the forEach loop) to avoid re-renders and leaks?

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,205 @@
...
+      const { status } = await axios.post<void>(
+        getMigrationOwnershipClaimsUrl(delegateId)
+      );
+      return { status };
+    } catch (error) {
+      return { status: -1 };
+    }
+  };
+
+  const fetchDelegateStatusPairs = () => {
```
gunzip, line 97: I'd make this method take `migrationDelegatesData` as input, see comment on useEffect

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,205 @@
...
+      "getServicesMigrationStatus",
+      {},
+      MigrationItemList,
+      { notify: "errors" }
+    );
+  };
+
+  // for each delegate fetch own migration status
+  useEffect(() => {
+    fetchDelegateStatusPairs();
```
gunzip, line 148:
```suggestion
    fetchDelegateStatusPairs(migrationDelegatesData);
```

make dependencies explict otherwise it gets difficult to track why `migrationDelegatesData` appears in the deps array an what other variables should.

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,205 @@
...
+      enqueueSnackbar(
+        buildSnackbarItem({
+          severity: "error",
+          title: t("notifications.exceptionError"),
+          message: JSON.stringify(error)
+        })
+      );
+    } finally {
+      setImportInProgress(false);
+      fetchMigrationLatestData();
```
gunzip, line 133: `fetchMigrationLatestData()` runs even when the above code fails, is it right?

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
```diff
@@ -0,0 +1,205 @@
+/**
+ * Azure Apim Subscription Manager
+ * Main component that manages the ownership change of Azure subscriptions from a user-centric model to an institution-centric model.
+ * This component manages:
+ * - viewing the current status of subscriptions migration (grouped by user)
+ * - the status update of this current view
+ * - the delegate selection modal for importing subscriptions into the Institution: this means that the ownership of the subscriptions,
+ *   once imported, passes from the individual user to the Institution of which these users are delegates.
+ */
```
gunzip, line 9:
```suggestion
/**
 * MigrationManager Component
 * 
 * The MigrationManager serves as the central interface for managing the migration of Azure subscriptions from individual user ownership to institutional ownership. This React component is part of the Azure Apim Subscription Manager and includes both UI and logic to handle various aspects of the migration process.
 * 
 * Capabilities:
 * - Displaying the current migration status of Azure subscriptions, grouped by user.
 * - Providing real-time updates on the migration progress.
 * - Allowing administrators to initiate and monitor the migration process.
 * - Facilitating the selection of delegates responsible for importing subscriptions into the institution, effectively transferring ownership from individual users to the institution.
 * 
 * Usage:
 * This component is intended for use within the administrative dashboard of the Azure Apim Subscription Manager, providing administrators with the tools needed to oversee and control the subscription migration process.
 */

```

gunzip: AI may help ;-)

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: Yes right, it's a porting of old Developer Portal Frontend logic

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: 23dfae6

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: Right, 23dfae6

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: a9aacf2

apps/backoffice/src/components/services/subscriptions-migration/migration-latest.tsx
giuseppedipinto: 1f277d8
you're absolutely right and I think I've always misunderstood this concept, thank you very much, I won't forget it!

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: `migrationDelegatesData` is the result of `useFetch` custom hook and for this reason I cannot use it "explicitly", at least in this context.
**P.S.:** Since I have an upcoming release deadline, I'm adding this issue as a technical debt task that I will evaluate asap.
Thanks for suggestions :-)

apps/backoffice/src/components/services/subscriptions-migration/migration-manager.tsx
giuseppedipinto: `migrationDelegatesData` is the result of `useFetch` custom hook and for this reason I cannot use it "explicitly", at least in this context.
**P.S.:** Since I have an upcoming release deadline, I'm adding this issue as a technical debt task that I will evaluate asap.
Thanks for suggestions :-)

gunzip (approved):
I'm approving to meet deadline but I'll try to suggest some refactors asap :-)

giamma1295 (approved):
All comments seems resolved or minor issues are traced as technical debts, so we can merge it

giuseppedipinto:
As I said in two similar comments above that I didn't follow up on, some things will be reviewed calmly after release.
In this first phase I was concerned with having a coherent and at least consistent operating logic.

@gunzip I hope you will surely appreciate, by watching the videos, _(almost always present in the frontend PR of important releases)_, that the user experience is absolutely dynamic and responsive (thanks to all the loading and feedback states).

There are certainly many ideas for improvement, many of which are already known and will be the subject of subsequent refactors and improvements.

gunzip:
Yep, good work ! ❤️  

What I'm trying to share here within comments are common principles that may be useful beyond this PR, ie.:

1. **documentation**: it is important to document the more obscure parts to make other devs aware about the _purpose_ of the business logic; usually a top level comment on each module suffices
2. **unnecessary complexity**: sometimes code is over engineered (ie. `useState`+`useEffect` vs simple variables), the less is better; most of the time when you think you need to introduce `useEffect` you don't (ie. you can directly put business logic into event handlers)
3. **maintanability**: think about modularity of the code, ie. avoiding giant components that mix business logic and UI (for example, using reducers, extracting methods from the component body, and so on)
4. **performance**: when developing frontend apps is important to not stress the browser too much, ie. avoiding too many re-renders in loops
5. **scalability**: when doing parallel calls consider backend capacity and throttling

Said so, I'm totally on the side of "if it works it works".

PR #499, opened 2023-11-08
chore: removed healtcheck which calls external services

giamma1295:
Remove healthchecks which was calling external services.

#### Checklist:


rodoviti (approved):
LGTM

PR #503, opened 2023-11-09
Service Preview Test

StefanoDoc:
Service Preview component unit tests
#### Checklist:

### DIFF
--- a/apps/backoffice/src/components/services/service-preview.tsx
+++ b/apps/backoffice/src/components/services/service-preview.tsx
@@ -44,10 +44,12 @@ export const ServicePreview = ({
 
   return (
     <Dialog open={isDialogOpen} onClose={handleClose} disableScrollLock>
-      <DialogTitle>{t("service.preview.title")}</DialogTitle>
+      <DialogTitle id="title">{t("service.preview.title")}</DialogTitle>
       <DialogContent sx={{ minWidth: "600px" }}>
         <Alert severity="info">
-          <Typography variant="body2">{t("service.preview.info")}</Typography>
+          <Typography id="info" variant="body2">
+            {t("service.preview.info")}
+          </Typography>
         </Alert>
         <Grid container marginY={4}>
           <Grid item xs>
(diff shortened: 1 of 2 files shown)


apps/backoffice/src/components/services/service-preview.tsx
```diff
@@ -44,10 +44,12 @@ export const ServicePreview = ({
 
   return (
     <Dialog open={isDialogOpen} onClose={handleClose} disableScrollLock>
-      <DialogTitle>{t("service.preview.title")}</DialogTitle>
+      <DialogTitle id="title">{t("service.preview.title")}</DialogTitle>
```
giamma1295, line 47: it is safe put as id a generic "title"? I mean can this id be erronuosly used in other elements cause is too generic and cause some issue?

apps/backoffice/src/components/services/service-preview.tsx
```diff
@@ -85,7 +87,7 @@ export const ServicePreview = ({
         </Grid>
       </DialogContent>
       <DialogActions sx={{ paddingX: 3, paddingY: 2 }}>
-        <Button variant="outlined" onClick={handleClose}>
+        <Button id="close-button" variant="outlined" onClick={handleClose}>
```
giamma1295, line 90: same as id

StefanoDoc: [9914d3f](https://github.com/pagopa/io-services-cms/pull/503/commits/9914d3f427282d776af4b7a8f9e6cbc08cfcec21)

apps/backoffice/src/components/services/service-preview.tsx
StefanoDoc: [9914d3f](https://github.com/pagopa/io-services-cms/pull/503/commits/9914d3f427282d776af4b7a8f9e6cbc08cfcec21)

giuseppedipinto (approved):

PR #514, opened 2023-11-10
fix issue on forwarding io-services-cms responses

giamma1295:
Workaround cause the generated client using codegen-ts does not map the response in case of status code !== 2XX/content-type !== application/json

#### Checklist:


apps/backoffice/mocks/handlers/services-cms-handlers.ts
```diff
@@ -64,126 +65,126 @@ export const buildHandlers = () => {
...
     }),
     rest.get(`${baseURL}/services/:serviceId/keys`, (_, res, ctx) => {
       const resultArray = [
         [ctx.status(200), ctx.json(getGetServiceKeys200Response())],
-        [ctx.status(401), ctx.json(null)],
-        [ctx.status(403), ctx.json(null)],
-        [ctx.status(404), ctx.json(null)],
-        [ctx.status(429), ctx.json(null)],
-        [ctx.status(500), ctx.json(null)]
+        [ctx.status(401), ctx.json(getIoServicesError(400))],
```
gunzip, line 106: why do we sometimes map each error code to the equivalent one, and in some other cases we just return 400 ?

apps/backoffice/mocks/handlers/services-cms-handlers.ts
```diff
@@ -64,126 +65,126 @@ export const buildHandlers = () => {
...
+        [ctx.status(429), ctx.json(getIoServicesError(400))],
+        [ctx.status(500), ctx.json(getIoServicesError(400))]
       ];
 
-      return res(...resultArray[0]);
+      return res(...resultArray[3]);
     }),
     rest.delete(`${baseURL}/services/:serviceId/release`, (_, res, ctx) => {
       const resultArray = [
         [ctx.status(204)],
```
gunzip, line 182: instead of having to map each return value from the backend, can't we just proxy the http call wihout doing this mapping?

PR #565, opened 2023-11-22
chore(fix): workaround on DELETED name received

giamma1295:
Handle legacy service having name strictly "DELETED" received on `OnLegacyServiceChange`
Unit Test

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
+++ b/apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
@@ -72,7 +72,7 @@ const fromLegacyToCmsService = (
       tos_url: service.serviceMetadata?.tosUrl,
       web_url: service.serviceMetadata?.webUrl,
     },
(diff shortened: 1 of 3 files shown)


giuseppedipinto (approved):

apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
```diff
@@ -192,3 +192,13 @@ export const handler =
...
     );
+
+const calculateServiceName = (serviceName: NonEmptyString) => {
+  const calculatedName = serviceName.replace("DELETED", "").trim();
+
+  if (NonEmptyString.is(calculatedName)) {
+    return calculatedName;
+  }
+
+  return "-" as NonEmptyString;
```
gunzip, line 203: another way is 

`return NonEmptyString.decode(calculatedName).getOrElse(() => "-")`

PR #570, opened 2023-11-23
bug: [IOPAE-824] Bring back institutions api

giamma1295:
Reintroduced GET /api/institutions API, which returns all the user's authorized institutions, in order to remove `authorizedInstitutions` field in session jwe token.
Needed in order to reduce JWE token lenght.
Unit Tests

#### Checklist:

### DIFF
--- a/apps/backoffice/mocks/handlers/selfcare-handlers.ts
+++ b/apps/backoffice/mocks/handlers/selfcare-handlers.ts
@@ -29,16 +29,7 @@ export const buildHandlers = () => {
     }),
     rest.get(`${baseURL}/institutions`, (_, res, ctx) => {
       const resultArray = [
-        [
-          ctx.status(200),
-          ctx.json(
-            [
-              ...Array.from(
-                Array(faker.number.int({ min: 140, max: 145 })).keys()
-              )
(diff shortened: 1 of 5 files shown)


apps/backoffice/mocks/handlers/selfcare-handlers.ts
```diff
@@ -29,7 +29,16 @@ export const buildHandlers = () => {
     }),
     rest.get(`${baseURL}/institutions`, (_, res, ctx) => {
       const resultArray = [
-        [ctx.status(200), ctx.json([getMockInstitution()])],
+        [
+          ctx.status(200),
+          ctx.json(
+            [
+              ...Array.from(
+                Array(faker.number.int({ min: 140, max: 145 })).keys()
```
rodoviti, line 37: **question:** are we sure we want `msw` to return all this authorized institutions?!

rodoviti (approved):
LGTM

apps/backoffice/mocks/handlers/selfcare-handlers.ts
giamma1295: leftover msw mock from a test, removed

PR #577, opened 2023-11-28
chore: Update B4F openapi definition: Replace bearer auth with cookie auth

giuseppedipinto:
Since the `next-auth` library is used to secure backend-frontend communication, the authentication is an encrypted cookie type _(as a the library standard)._

The _openapi_ specification was therefore cleaned up and updated.
The same thing was done for the `useFetch` react hook used by the frontend for communication with the backend.

#### List of Changes
- updated `openapi` **securitySchemes** definition
- updated `useFetch` hook _(cleanup)_
- cleanup unused `/auth` api _(and related mocks)_

Local `dev` mode with `msw` mocks

#### Checklist:


rodoviti (approved):
LGTM

apps/backoffice/openapi.yaml
```diff
@@ -11,12 +11,10 @@ info:
   contact:
     name: PagoPA S.p.A.
     url: https://docs.pagopa.it/io-guida-tecnica/
-  version: 0.2.0
+  version: 1.0.0
 servers:
   - url: http://localhost:3000/api #FIXME
```
giamma1295, line 16: We have this value?

PR #592, opened 2023-12-06
[#IOPAE-869] Avoid Checking Authorized CIDRs in B4F Request forwarded to CMS

giamma1295:
For requests coming from Backoffice B4F no check for authorized CIDRs needs to be done.
#### List of Changes
- Added `AzureUserAttributesManageMiddlewareWrapper`, a Middleware Wrapper for `AzureUserAttributesManageMiddleware`, this wrapper executes the original middleware, but will returns an empty Authorized CIDRs list when the request comes from the Backoffice B4F subnet, this in order to skip the authorized CIDRs tests in all controller.
- Added `AzureUserAttributesManageMiddlewareWrapper` Unit Tests
- Updated Existing Controllers Unit Test 

Unit Tests

#### Checklist:


apps/io-services-cms-webapp/src/utils/azure-user-attributes-manage-middleware-wrapper.ts
```diff
@@ -0,0 +1,57 @@
+import { EmailAddress } from "@pagopa/io-functions-commons/dist/generated/definitions/EmailAddress";
+import { SubscriptionCIDRsModel } from "@pagopa/io-functions-commons/dist/src/models/subscription_cidrs";
+import {
+  AzureUserAttributesManageMiddleware,
+  IAzureUserAttributesManage,
+} from "@pagopa/io-functions-commons/dist/src/utils/middlewares/azure_user_attributes_manage";
+import { IRequestMiddleware } from "@pagopa/ts-commons/lib/request_middleware";
+import { IResponse } from "@pagopa/ts-commons/lib/responses";
+import * as E from "fp-ts/lib/Either";
+import CIDR from "ip-cidr";
```
gunzip, line 10: if you use `io-functions-commons` you probably already have `cidr-matcher` in your deps. 

https://github.com/pagopa/io-functions-commons/blob/74738573db3b036fbcde2c1e169ae647c65351fb/package.json#L61

giamma1295: Thank's, replaced ip-cidr with cidr-matcher!

giuseppedipinto (approved):

PR #609, opened 2023-12-13
[IOPAE-793] Frontend topic management on create/update service

giuseppedipinto:
Added **topic** management when **creating/editing** a service.
Service topic is an optional property that can be set in the create/update form.

#### List of Changes
- update backoffice `openapi` with topic data
- update Service **types**, **adapters** and all topic **related data**
- update **translations**
- update `MSW` mocks with topic data
- add new `AutocompleteController` with **MUI** `Autocomplete` component
- add **frontend logic** to manage service topic on create/update form
- minor adjustments

#### List of Changes (update)
- c90f589 fix MUI warning _(MUI: The value provided to Autocomplete is invalid. None of the options match with `""`.)_
- 35975fc update topic field layout: reduce width to half

#### Motivation and Context
Need to manage the topic of a service

Unit tests
Local `dev` mode with `msw` mocks

https://github.com/pagopa/io-services-cms/assets/118166285/b00f27f1-ee8c-40b2-bdf5-b9e44d4ce2cc

Updated layout:

https://github.com/pagopa/io-services-cms/assets/118166285/d176b9c8-9aed-4b5c-9aa7-53f3247d538b

#### Checklist:

### DIFF
--- a/apps/backoffice/src/components/services/adapters.ts
+++ b/apps/backoffice/src/components/services/adapters.ts
@@ -96,40 +96,38 @@ export const fromServiceCreateUpdatePayloadToApiServicePayload = (
   );
 
 /**
- * Convert `@/generated/api/ServicePayload` to _'frontend'_ service payload
- * @param apiService service _(as result api getService fetch)_
- * @param sessionUser `next-auth` session user
+ * Convert `@/generated/api/ServiceLifecycle` to _'frontend'_ service payload
+ * @param sl `ServiceLifecycle` _(as result of api getService fetch)_
  * @returns `ServiceCreateUpdatePayload`
  */
-export const fromApiServicePayloadToServiceCreateUpdatePayload = (
-  apiService: ApiServicePayload
+export const fromServiceLifecycleToServiceCreateUpdatePayload = (
+  sl: ServiceLifecycle
 ): ServiceCreateUpdatePayload => ({
-  name: apiService.name,
-  description: apiService.description,
-  require_secure_channel: apiService.require_secure_channel ?? false,
-  authorized_cidrs: apiService.authorized_cidrs
-    ? Array.from(apiService.authorized_cidrs.values())
+  name: sl.name,
+  description: sl.description,
+  require_secure_channel: sl.require_secure_channel ?? false,
+  authorized_cidrs: sl.authorized_cidrs
+    ? Array.from(sl.authorized_cidrs.values())
     : [],
-  authorized_recipients: apiService.authorized_recipients
-    ? Array.from(apiService.authorized_recipients.values())
+  authorized_recipients: sl.authorized_recipients
+    ? Array.from(sl.authorized_recipients.values())
     : [],
-  max_allowed_payment_amount: apiService.max_allowed_payment_amount ?? 0,
+  max_allowed_payment_amount: sl.max_allowed_payment_amount ?? 0,
   metadata: {
-    web_url: apiService.metadata.web_url ?? "",
-    app_ios: apiService.metadata.app_ios ?? "",
-    app_android: apiService.metadata.app_android ?? "",
-    tos_url: apiService.metadata.tos_url ?? "",
-    privacy_url: apiService.metadata.privacy_url ?? "",
-    address: apiService.metadata.address ?? "",
-    assistanceChannels: convertAssistanceChannelsObjToArray(
-      apiService.metadata
-    ),
-    cta: buildCtaObj(apiService.metadata.cta),
-    token_name: apiService.metadata.token_name ?? "",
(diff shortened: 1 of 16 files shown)


apps/backoffice/src/components/services/adapters.ts
```diff
@@ -228,11 +227,11 @@ const getCtaUrlFromCtaString = (value: string) => {
  * or on future uses of such fields which may be undefined.
  * @param s service
  * @returns service without epmpty or undefined properties */
-const clearUndefinedEmptyProperties = (s: any): any => {
+const clearUndefinedNullEmptyProperties = (s: any): any => {
```
gunzip, line 230: this is the only method in this repo that requires lodash. lodash provides the same features as fp-ts in a more imperative fashion. I strongly suggest to remove lodash in order to reduce the size of the bundle and rewrite this method. You can use plain javascript:

```ts
/** Remove temporary assistanceChannels[] metadata used for react-hook-form dynamic array fields. */
const removeAssistanceChannelsArray = (s: any) => {
  const { assistanceChannels, ...metadata } = s.metadata;
  return { ...s, metadata };
};

type DeepPartial<T> = {
  [P in keyof T]?: DeepPartial<T[P]>;
};

function clearUndefinedNullEmptyProperties<T extends object>(
  obj: T,
  filterOutValues = new Set([null, undefined, ""])
): DeepPartial<T> {
  return Object.entries(obj).reduce((acc, [key, value]) => {
    if (filterOutValues.has(value)) {
      return acc;
    }

    if (typeof value === "object" && !Array.isArray(value)) {
      return { ...acc, [key]: clearUndefinedNullEmptyProperties(value) };
    }

    return { ...acc, [key]: value };
  }, {});
}

```

gunzip (changes requested):

apps/backoffice/src/components/services/adapters.ts
giuseppedipinto: Hi @gunzip, I actually use lodash also in another component _(for object deep comparison)_.
In general I consider `lodash` a very valid _(and widely used)_ library. 
Furthermore, IMO, it is convenient and powerful as it expresses concepts in a minimal way.
However, I accept your objection and I have opened an internal technical debt [task](https://pagopa.atlassian.net/browse/IOPAE-902) to evaluate the problem globally on the repo.

gunzip: Can you share a link with the code where lodash is used?

giuseppedipinto: https://github.com/pagopa/io-services-cms/blob/b1e6dec1059b7b75b02c96c667e37ade2ac2b91e/apps/backoffice/src/components/create-update-process/create-update-process.tsx#L103

gunzip: You can use something alike

```ts
export const isDeepEqual = (a: any, b: any): boolean => {
  if (a === b) {
    return true;
  }

  if (typeof a !== typeof b) {
    return false;
  }

  if (typeof a !== "object" || a === null || b === null) {
    return false;
  }

  if (Object.keys(a).length !== Object.keys(b).length) {
    return false;
  }

  if (Array.isArray(a) !== Array.isArray(b)) {
    return false;
  }

  if (Array.isArray(a)) {
    return a.every((item, index) => isDeepEqual(item, b[index]));
  }

  return Object.keys(a).every(key => isDeepEqual(a[key], b[key]));
};

```

or

```ts
export const isDeepEqual = (a: any, b: any): boolean => {
  return JSON.stringify(a) === JSON.stringify(b);
};

```

giuseppedipinto: Sorry but it's not the same.
json.stringify assumes that the strings are ordered in the same way _(too ugly to look at! )_ while in the other case it doesn't seem acceptable to me to write about 20 lines of codes _(which also have to be tested)_ to do a job that I can achieve with one line...it seems more like a trial of intentions than a real need to eliminate a library.
Anyway we will evaluate this possibility in the technical debt task I reported above, but I would like to let you know that I do not agree with the approach

gunzip: In which case 20 lines of code are worse than serving an entire library just to use 2 methods? Furthermore, consider in your evalutation that 80% of lodash features are covered by fp-ts which is a library you're already using.

giuseppedipinto: I will evaluate it in internal technical debt [task](https://pagopa.atlassian.net/browse/IOPAE-902).
Thank you so much for the valuable advice and food for thought, as always 🙏😉

gunzip (approved):

StefanoDoc (approved):

PR #628, opened 2023-12-20
[#IOPAE-901] service topic list implementation

giamma1295:
Implements Retrive All Service Topics API.

#### List of Changes
- Add `function.json` for `GetServiceTopics` AZF Http Handler (GET /services/topics)
- Implements GET /services/topics API
- Add Unit Tests for GET /services/topics API

Unit Tests

#### Checklist:

- [ ] Add APIM cache policy to avoid execute query on every request


apps/io-services-cms-webapp/src/webservice/controllers/get-service-topics.ts
```diff
@@ -0,0 +1,96 @@
...
+    telemetryClient,
+    serviceTopicDao,
+  }: Dependencies): GetServiceTopicsHandler =>
+  (context, auth) =>
+    pipe(
+      serviceTopicDao.findAllNotDeletedTopics(),
+      TE.map(
+        flow(
+          O.fold(
+            // QUESTION: when no topics are found on DB should we return an empty list or an error 404?
```
rodoviti, line 58: **note:** in general, for rest services like "findAll", "search" or something like that, I prefer to return an empty list rather that a 404, but honestly I think both solutions are valid, so is up to us to choose the one we prefer

rodoviti (approved):
**thought:** I think we should review (separately from this development), the dependency injection strategy used, because the fact that adding a new dependency used by only 1 controller impacts the unit tests of all other controllers, imho does not seem to be a good thing

apps/io-services-cms-webapp/src/webservice/controllers/get-service-topics.ts
gunzip: I agree! It should return an empty lists, not a 404.

Ideally REST endpoints that returns array should also wrap theme inside object (in other words, an object should always be the root data stracture), this approach lets you add metadata (ie. legth, page, ...) and is future proof. Something like:

```
{
  items: [ ... ],
}
```

not need for this PR anyway.

PR #630, opened 2024-01-02
fix: [#IOPAE-912] replace findOneByServiceId method call with findLastVersionByModelId

giamma1295:
- call `findLastVersionByModelId` instead of deprecated `findOneByServiceId` method
#### Checklist:


rodoviti (approved):
LGTM

PR #653, opened 2024-01-12
[#IOPAE-904] First service activation info on review ticket

giamma1295:
Added in the ticket if the review request is related to a first service activation or a subsequent change.
Unit Test

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/utils/__tests__/jira-proxy.test.ts
+++ b/apps/io-services-cms-webapp/src/utils/__tests__/jira-proxy.test.ts
@@ -107,11 +107,11 @@ describe("Service Review Proxy", () => {
 
     const aJiraClient: JiraAPIClient = jiraClient(JIRA_CONFIG, mockFetch);
     const proxy = jiraProxy(aJiraClient);
-    const firstReview = true;
+    const firstPublication = true;
     const serviceReviewTicket = await proxy.createJiraIssue(
       aService,
       aDelegate,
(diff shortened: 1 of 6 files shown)


apps/io-services-cms-webapp/src/utils/__tests__/jira-proxy.test.ts
```diff
@@ -107,9 +107,11 @@ describe("Service Review Proxy", () => {
 
     const aJiraClient: JiraAPIClient = jiraClient(JIRA_CONFIG, mockFetch);
     const proxy = jiraProxy(aJiraClient);
+    const firstReview = true;
```
rodoviti, line 110:
```suggestion
    const firstPublication = true;
```

apps/io-services-cms-webapp/src/utils/__tests__/jira-proxy.test.ts
```diff
@@ -107,9 +107,11 @@ describe("Service Review Proxy", () => {
...
     const serviceReviewTicket = await proxy.createJiraIssue(
       aService,
-      aDelegate
+      aDelegate,
+      firstReview
```
rodoviti, line 114:
```suggestion
      firstPublication
```

rodoviti (commented):
I would "standardize" the variable name.
Everything else is ok for me

rodoviti (approved):

PR #664, opened 2024-01-16
[#IOPAE-932] Sync on legacy when service was never published

giamma1295:
Sync on legacy when service was never published.
Unit Tests

#### Checklist:


rodoviti (approved):
LGTM

PR #672, opened 2024-01-19
[IOPAE-934] Retrieve service topics backoffice

giamma1295:
Retrieve Service Topics on Backoffice.
#### List of Changes
- Implements TopicsProvider, this component retrieve and cache the service topics list on backoffice calling io-services-cms webb API.

Unit Tests

#### Checklist:


apps/backoffice/src/lib/be/cms-client.ts
```diff
@@ -48,6 +56,61 @@ export const getIoServicesCmsClient = (): Client => {
...
+      throw new Error("blank response");
+    }
+
+    // caching topics for future requests
+    // expiration is set to now + API_SERIVCES_CMS_TOPICS_CACHE_TTL_MINUTES config
+    const now = new Date();
+    now.setMinutes(
+      now.getMinutes() + API_SERIVCES_CMS_TOPICS_CACHE_TTL_MINUTES
+    );
+    cachedServiceTopicsExpiration = now;
```
rodoviti, line 96: **suggestion (non-blocking):** I think it's clearer that way
```suggestion
    cachedServiceTopicsExpiration = now.getMinutes() + API_SERIVCES_CMS_TOPICS_CACHE_TTL_MINUTES;
```

apps/io-services-cms-webapp/openapi.yaml
```diff
@@ -132,12 +132,6 @@ paths:
       summary: Retrieve all services
       description: Retrieve all service topics
       operationId: getServiceTopics
-      parameters:
-        - $ref: '#/components/parameters/UserEmail'
-        - $ref: '#/components/parameters/UserGroups'
-        - $ref: '#/components/parameters/UserID'
-        - $ref: '#/components/parameters/SubscriptionID'
-        - $ref: '#/components/parameters/XForwardedFor'
```
rodoviti, line 140: **chore:** from the _Services CMS_ point of view, by removing _XForwardedFor_ we "lost" the origin caller IP....Azure will automatically add to this header only the last caller IP (i.e. the _Backoffice_ IP)

rodoviti, line 140: chore: from the _Services CMS App_ point of view, by removing _XForwardedFor_ (i.e.: avoiding to explicitly setting _XForwardedFor_ header) we "lost" the origin caller IP....Azure will automatically add to this header only the last caller IP (i.e. the _Backoffice App_ IP)

giamma1295: added in 07afcb591d3b9e39a6ec23067d6ae333b1c05be7

rodoviti (approved):
LGTM

PR #711, opened 2024-02-01
[#IOPAE-955] Sanitize Route Handlers response objects

giamma1295:
Sanitize response payload in order to prevent XSS.
#### List of Changes
- Adding `xss` dependency to `backoffice`
- Adding `sanitizedNextResponseJson` function,a function which build a NextResponse with a sanitized response object

#### Checklist:

### DIFF
--- a/apps/backoffice/src/lib/be/__tests__/sanitize.test.ts
+++ b/apps/backoffice/src/lib/be/__tests__/sanitize.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it } from "vitest";
+import xss from "xss";
+import { sanitizeObject } from "../sanitize";
+
+const aService = {
+  id: "aServiceId",
+  status: {
+    value: "submitted"
+  },
+  last_update: "2024-01-31T14:44:57.000Z",
+  metadata: {
+    privacy_url: "https://testprivacy.it",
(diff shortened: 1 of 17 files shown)


apps/backoffice/src/lib/be/__tests__/sanitize.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(result).not.toEqual(anUnsafeService);
+    expect(result).toStrictEqual({
+      ...anUnsafeService,
+      authorized_recipients: [
+        "<img>",
+        "AAAAAA00A00A000A",
+        "<img>",
+        "&lt;script&gt;alert('You Have been Pwned!')&lt;/script&gt;"
+      ]
+    });
```
StefanoDoc, line 98: how about a case like this ? ->

giamma1295: yes! as shown on e7d2b0e239d35f2abee01d6496b7ae5dd8d601d9

giamma1295: xss library checks prevent running malicious code and threat by discover and remove it

StefanoDoc (approved):

PR #742, opened 2024-02-19
[#IOPAE-968] preserve topic_id on sync legacy

giamma1295:
Handle `topic_id` in Legacy -> CMS sync process.
#### List of Changes
- Preserve current topic_id both for ServiceLifecycle and ServicePublication sync items when present.
- Fallback to the ServiceLifecycle `topic_id` on ServicePublication sync and no item is found on the ServicePublication collection (this happen when the first publication is done on  legacy Developer Portal)
- Use the Default `topic_id` when this service is not found on Both ServiceLifecycle and ServicePublication collections.

Unit Tests

#### Checklist:

### DIFF
--- a/.changeset/shaggy-frogs-shave.md
+++ b/.changeset/shaggy-frogs-shave.md
@@ -1,0 +1,5 @@
+---
+"io-services-cms-webapp": patch
+---
+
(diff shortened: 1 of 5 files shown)


rodoviti (approved):
LGTM

PR #747, opened 2024-02-20
create readonly user migrations scripts

rodoviti:
Add a new user to Postgres DB with only read permission

#### List of Changes

Add a flyway migration scripts

#### Motivation and Context

To allow other teams (such as Operation, Service Management, etc) to retrieve/monitor data

#### Checklist:


giuseppedipinto (approved):
LGTM

PR #748, opened 2024-02-20
WIP: first handler draft

giamma1295:
#### Checklist:


apps/io-services-cms-webapp/src/main.ts
```diff
@@ -2,6 +2,7 @@ import { ApimUtils } from "@io-services-cms/external-clients";
 import {
   LegacyService,
   ServiceHistory,
+  ServiceHistory as ServiceHistoryCosmosType,
```
rodoviti, line 5: **question**: why are there two imports of the same type?

giamma1295: typo, fixed in 4690baa4301f6215b74dec93e6673ac5cb3c055d

PR #764, opened 2024-03-01
[IOPAE-1009] Zendesk webform integration

giuseppedipinto:
**_Zendesk webform integration by implementing solution 2, i.e. frontend integration._**

This integration method does not involve the invocation of a specific API but rather a simple redirect to the Reserved Area support page.
In particular, the IO Backoffice redirects the user to the following link `ENV.selfcare.pagopa.it/associazione?productId=productId`.
The value of the query parameter `productId` corresponds to the identifier of the product being integrated, i.e. IO _(prod-io)_.

#### List of Changes
- remove old implementation _(a backend integration)_
- add new implementation _(frontend only integration)_

Local `dev` mode with `msw` mocks.

#### Checklist:


apps/backoffice/src/lib/be/institutions/business.ts
```diff
@@ -1,14 +1,8 @@
-import { AssistanceItem } from "@/generated/api/AssistanceItem";
-import { AssistanceResponse } from "@/generated/api/AssistanceResponse";
 import { Institution as BackofficeInstitution } from "@/generated/api/Institution";
 import { UserAuthorizedInstitution } from "@/generated/api/UserAuthorizedInstitution";
 import { UserAuthorizedInstitutions } from "@/generated/api/UserAuthorizedInstitutions";
 import { InstitutionResource } from "@/types/selfcare/InstitutionResource";
-import {
-  getInstitutionById,
-  getUserAuthorizedInstitutions,
-  sendSupportRequest
```
giamma1295, line 10: Is the method `sendSupportRequest` removed also on selfcare-client.ts?

giuseppedipinto: de2731c9 _(I was forgetting it, thanks)_

giamma1295 (approved):
LGTM

PR #783, opened 2024-03-19
[IOPAE-666] Physically delete service on service-publication

giamma1295:
With this pr we implement the **physical deletion of a service from the service-publication cosmosDB container** following a service-deletion action.

BTW the service remains in the cosmosDB service-lifecycle container(with status `deleted`, logical deletion), in addition the deleted service-publication entry remains replicated on the service-history container(consequently it will be possible to restore, with manual intervention, the deleted service).

#### List of Changes
- Add new Azure Function `OnServiceDeletion`, this QueueTrigger(listen on `request-deletion` queue) type function will evetually delete received services fron `service-publication` container.
- Updated `LegacyServiceWatcher` Azure Function, in order to stop creating an unpublished record when a legacy service deletion is intercepte but only a delete record for sync.
- Update `ServiceLifecycleWatcher` Azure Function in order to intercept deletetion events, both from `io-service-cms` and `legacy`, and write a message on `request-deletion` queue
- Add `delete` operation to FSMStore, both memory and cosmos
- Updated unit tests

#### Motivation and Context
Fix an edge case which is allowing a deleted service to be published.

Unit test and Local Application Run

#### Checklist:


apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
```diff
@@ -65,58 +65,66 @@ const onApproveHandler =
...
-    }
-  };
+  ({ item }) =>
+    pipe(
+      item,
+      onAnyChangesHandler,
+      E.right,
+      E.chain((historicizationAction) =>
+        pipe(
+          item.fsm.lastTransition !== SYNC_FROM_LEGACY ||
```
gunzip, line 113: can you add a comment?

apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
```diff
@@ -65,58 +65,66 @@ const onApproveHandler =
...
+            item.fsm.state === "deleted",
+          B.fold(
+            () => E.right(historicizationAction),
+            () =>
+              pipe(
+                item,
+                getSpecificAction(config),
```
gunzip, line 120: ditto

packages/io-services-cms-models/src/lib/__tests__/store.cosmos.test.ts
```diff
@@ -241,4 +241,54 @@ describe("store cosmos tests", () => {
       expect(result.right.version).toBe(anEtag);
     }
   });
+
+  it("shold delete the element", async () => {
```
gunzip, line 245:
```suggestion
  it("should delete the element", async () => {
```

packages/io-services-cms-models/src/lib/fsm/store.cosmos.ts
```diff
@@ -126,5 +127,37 @@ export const createCosmosStore = <
       }))
     );
 
-  return { fetch, bulkFetch, save };
+  // https://learn.microsoft.com/en-us/rest/api/cosmos-db/delete-a-document
+  // expected status code return are:
+  // - 204: The documenti was successfully deleted.
+  // - 404: The documenti was not found.
```
gunzip, line 133:
```suggestion
  // - 204: The document was successfully deleted.
  // - 404: The document was not found.
```

apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
giamma1295: 2d32e4c8978af5f93573a9c52ff4768d6692b4cc

apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
giamma1295: 2d32e4c8978af5f93573a9c52ff4768d6692b4cc

packages/io-services-cms-models/src/lib/__tests__/store.cosmos.test.ts
giamma1295: d1706a746a439ca36a11a2bd465f96c14f2254f2

packages/io-services-cms-models/src/lib/fsm/store.cosmos.ts
giamma1295: 8e40c603ee75d47f13ca1dcb563fec36991daeda

rodoviti (approved):
LGTM

PR #805, opened 2024-04-17
[#IOPAE-1079] PaginationResultSet mandatory parameters

giamma1295:
Change PaginationResultSet openapi schema, set all parameters mandatory
#### Checklist:

### DIFF
--- a/packages/io-services-cms-models/openapi/commons-schemas.yaml
+++ b/packages/io-services-cms-models/openapi/commons-schemas.yaml
@@ -5,7 +5,7 @@ Cidr:
 FiscalCode:
   type: string
   description: User's fiscal code.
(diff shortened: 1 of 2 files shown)


packages/io-services-cms-models/openapi/commons-schemas.yaml
```diff
@@ -34,6 +34,10 @@ PaginationResultSet:
     count:
       type: number
       description: total record count
+  reuqired:
```
giuseppedipinto, line 37: typo?

giuseppedipinto (changes requested):

packages/io-services-cms-models/openapi/commons-schemas.yaml
giamma1295: 1b2245980f189ab5e40ce83d198e6ced380a5019

giuseppedipinto (approved):

PR #807, opened 2024-04-17
EC-234 - Implemented network and ai search resources

christian-calabrese:
#### List of Changes

Implemented network and AI search resources in Italy North region.
Employed the new terraform repository structure introduced by @Krusty93 in the io-infra Italy North region.

#### Motivation and Context

The io-services-cms domain has the need to implement the free-text search in the IO app. This PR introduces the needed infrastructure resources to initialize a subnet and AI Search related resources in the Italy North region.

Changes are only relative to the infrastructure. Therefore, precommit checks and plans have been tested succesfully.

#### Checklist:

### DIFF
--- a/infra/prod/italynorth/locals.tf
+++ b/infra/prod/italynorth/locals.tf
@@ -1,0 +1,19 @@
+locals {
+  prefix               = "io"
+  env_short            = "p"
+  location_short       = "itn"
+  location             = "italynorth"
+  project              = "${local.prefix}-${local.env_short}-${local.location_short}"
+  is_prod              = local.env_short == "p" ? true : false
+  application_basename = "services-cms"
+  srch_snet_cidrs      = ["10.20.135.0/26"]
+
+  tags = {
+    CreatedBy   = "Terraform"
+    Environment = "Prod"
+    Owner       = "IO"
+    Source      = "https://github.com/pagopa/io-services-cms"
+    CostCenter  = "TS310 - PAGAMENTI & SERVIZI"
+  }
+}
+
--- a/infra/_modules/ai_search/.terraform.lock.hcl
+++ b/infra/_modules/ai_search/.terraform.lock.hcl
@@ -1,0 +1,24 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/azurerm" {
+  version = "3.99.0"
+  hashes = [
+    "h1:1+d7Ciq/7GN4hY/+VshO2p4uOlUqHH6WpK2Zu4YocqE=",
+    "h1:b24Yw8/EneYHRV3aPoVexmw0Eo252ur30tN+sPajzXk=",
+    "h1:dawmYJUMGlL3t1mKDyaLJc08uSxPaUBoCAb/YCbVxPM=",
+    "h1:yHNaEhlR3kqlItAXFLWlIH2xxu4i7r2XzQnS04f/qBo=",
+    "zh:20581c1f4c586a37af45ed4c2a86ff4d868cee79139a755bd29750d804cee3ef",
+    "zh:28b3cc4e5f8bc65a595eab011d5965203a39e92aa9e26df842ffc979305ac823",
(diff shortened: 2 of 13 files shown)


infra/prod/italynorth/locals.tf
```diff
@@ -0,0 +1,19 @@
...
+  location             = "italynorth"
+  project              = "${local.prefix}-${local.env_short}-${local.location_short}"
+  is_prod              = local.env_short == "p" ? true : false
+  application_basename = "services-cms"
+  srch_snet_cidrs      = ["10.20.135.0/26"]
+
+  tags = {
+    CreatedBy   = "Terraform"
+    Environment = "Prod"
+    Owner       = "IO"
```
Krusty93, line 14: ManagementTeam = "IO Enti & Servizi"

Krusty93 (approved):
LGTM

infra/prod/italynorth/locals.tf
```diff
@@ -0,0 +1,19 @@
+locals {
+  prefix               = "io"
+  env_short            = "p"
+  location_short       = "itn"
+  location             = "italynorth"
+  project              = "${local.prefix}-${local.env_short}-${local.location_short}"
+  is_prod              = local.env_short == "p" ? true : false
+  application_basename = "services-cms"
```
rodoviti, line 8: suggestion (blocking): keep `application_basename` more generic

@giuseppedipinto @giamma1295 @StefanoDoc do you agree to keep `application_basename` more generic and add, for example, `-cms` or `-app-be` to specific resources?
I mean, the resources in this PR are not related to Services CMS, so I think we should not add `-cms` suffix to them to keep their names clean.
Maybe, we can add two more vars (i.e. `application_basename_cms` and `application_basename_app`) composed by `application_basename.concat("<specific_suffix>")`
```suggestion
  application_basename = "services"
```

giamma1295: I agree, services-cms is too specific for a single application in this monorepo, services or io-services would be better, however the repo name is "io-services-cms", so we should rename it to tend as well, to somenthing more generic like "io-services"

giamma1295 (approved):
LGTM

.pre-commit-config.yaml
```diff
@@ -8,8 +8,8 @@ repos:
         files: ^infra/
         args:
           - --args=--exclude-downloaded-modules
-      - id: terraform_validate
-        args:
-          - --init-args=-lockfile=readonly
-          - --args=-json
-          - --args=-no-color
+      # - id: terraform_validate
```
gunzip, line 11: why this change?

infra/prod/italynorth/locals.tf
```diff
@@ -0,0 +1,19 @@
+locals {
+  prefix               = "io"
+  env_short            = "p"
+  location_short       = "itn"
+  location             = "italynorth"
+  project              = "${local.prefix}-${local.env_short}-${local.location_short}"
+  application_basename = "services"
+  srch_snet_cidrs      = ["10.20.135.0/26"]
```
gunzip, line 8: how did you get this value?

.pre-commit-config.yaml
christian-calabrese: The newly introduced repository structure with modules and a subfolder <environment>/<region>/ is causing this pre-commit check to fail. 

This check appears to be flagging the different versions of the azurerm provider requested by each module, which imho contradicts the principle of reusable modules.

Together with @Krusty93, we are investigating the issue, but no solution has been found yet. The idea is to enable it back when a solution is found. What do you think?

infra/prod/italynorth/locals.tf
christian-calabrese: With the aim of introducing a CIDR allocation strategy in the near future, we've opted for a similar IP class to the one employed by the project in the west Europe region (10.0.135.*/26).

For the other values, instead, I took inspiration from the work done by @Krusty93 when implementing the [IaC for io-infra Italy North region](https://github.com/pagopa/io-infra/tree/main/src/core/prod/italynorth).

rodoviti (approved):
LGTM

.pre-commit-config.yaml
gunzip: it's ok, but I'd add this information as a short comment inside.

PR #812, opened 2024-04-23
EC-234 - Implemented cicd for prod italynorth infrastructure

christian-calabrese:
#### List of Changes
Added CI and CD to deploy the prod italy north infrastructure

#### Checklist:

### DIFF
--- a/.github/workflows/infra-cd.yml
+++ b/.github/workflows/infra-cd.yml
@@ -1,4 +1,4 @@
-name: Terraform apply on prod weu infra
+name: Terraform apply weu
 
 on:
   workflow_dispatch:
--- a/.github/workflows/infra-cd.yml
+++ b/.github/workflows/infra-cd.yml
@@ -1,4 +1,4 @@
-name: Run terraform apply on prod westeurope infrastructure
+name: Terraform apply on prod weu infra
 
 on:
   workflow_dispatch:
--- a/.github/workflows/infra-cd.yml
+++ b/.github/workflows/infra-cd.yml
@@ -1,4 +1,4 @@
-name: Infra Continuous Delivery
+name: Infra Continuous Delivery Prod
 
 on:
   workflow_dispatch:
--- a/.github/workflows/infra-cd.yml
+++ b/.github/workflows/infra-cd.yml
@@ -1,4 +1,4 @@
-name: Infra Continuous Delivery Prod
+name: Run terraform apply on prod westeurope infrastructure
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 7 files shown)


.github/workflows/infra-release.yml
```diff
@@ -0,0 +1,22 @@
...
+      - master
+    paths:
+      - "infra/prod/italynorth/**"
+      - ".github/workflows/infra-review.yml"
+      - ".github/workflows/infra-release.yml"
+
+jobs:
+  infra_release_prod_itn:
+    uses: pagopa/dx/.github/workflows/infra_apply.yaml@2a6b15a40418406315ae881d3c5c5c1988630ae1
+    name: Release Infrastructure Prod Italy North
```
Krusty93, line 16: I'd set `Prod Italy North` because GitHub concatenates the workflow name, the job name and the workflow call's job name

.github/workflows/infra-review-p-itn.yml
```diff
@@ -0,0 +1,25 @@
+name: PR Infrastructure Prod Italy North
```
gunzip, line 1: can we change these titles with the intent of these workflow? ie. something alike
"Run Terraform appy on..." "Run Terraform plan on..." 

it's not clear to me why we have 4 workflows for infra

.github/workflows/infra-cd.yml
```diff
@@ -1,4 +1,4 @@
-name: Infra Continuous Delivery
+name: Infra Continuous Delivery Prod
```
gunzip, line 1: what does this workflow?

.github/workflows/infra-cd.yml
```diff
@@ -1,4 +1,4 @@
-name: Infra Continuous Delivery
+name: Run terraform apply on prod westeurope infrastructure
```
Krusty93, line 1: If you agree, I'd use a short name because of the way GitHub UI is designed. Example `Release Infra weu`

christian-calabrese: @gunzip suggested to explicitly explain the operation being performed (plan or apply). What do you think about Terraform Plan weu and Terraform Apply weu?

gunzip: Is there any need to specify the region?

christian-calabrese: Currently, the workflows that execute plan and apply are different for the two regions where the infrastructure is deployed (itn and weu). I think that adding the region gives additional clarity on which region is being deployed.

giamma1295 (approved):

PR #815, opened 2024-04-23
EC-234 - Indices, datasources and indexers implemented 

christian-calabrese:
#### List of Changes
Indices, datasources and indexers implemented and minor fixes to networking

#### Motivation and Context
Apart from the creation of mandatory ai search resources (indices, indexers and datasources), the networking changes are the following:
- Moved the ai search private endpoint to a dedicated common subnet
- Changed the cidr to fill the vnet block from the lowest class

Terraform plan

#### Checklist:

### DIFF
--- a/infra/_modules/ai_search/providers.tf
+++ b/infra/_modules/ai_search/providers.tf
@@ -8,9 +8,10 @@ terraform {
 }
 
 provider "restapi" {
-  uri                  = "https://${azurerm_search_service.srch.name}.search.windows.net"
+  uri                  = "https://${azurerm_search_service.srch.name}.privatelink.search.windows.net"
   write_returns_object = true
   debug                = true
+  insecure             = true # Avoid certificate checking since it's a vnet internal resource
 
   headers = {
     "api-key"      = azurerm_search_service.srch.primary_key,
--- a/infra/_modules/ai_search/providers.tf
+++ b/infra/_modules/ai_search/providers.tf
@@ -1,0 +1,23 @@
+terraform {
+  required_providers {
+    restapi = {
+      source  = "Mastercard/restapi"
+      version = "<= 1.19.1"
+    }
+  }
+}
+
+provider "restapi" {
+  uri                  = "https://${azurerm_search_service.srch.name}.search.windows.net"
+  write_returns_object = true
+  debug                = true
+
+  headers = {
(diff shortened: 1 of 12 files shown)


infra/_modules/ai_search/providers.tf
```diff
@@ -0,0 +1,23 @@
+terraform {
+  required_providers {
+    restapi = {
+      source  = "Mastercard/restapi"
```
gunzip, line 4: I think we cannot use these providers from third party vendors. @Krusty93 what about this?

infra/prod/italynorth/locals.tf
```diff
@@ -5,7 +5,7 @@ locals {
   location             = "italynorth"
   project              = "${local.prefix}-${local.env_short}-${local.location_short}"
   application_basename = "services"
-  srch_snet_cidrs      = ["10.20.135.0/26"]
+  srch_snet_cidrs      = ["10.20.4.0/26"]
```
gunzip, line 8: I'd add a comment about where this value come from

infra/prod/italynorth/providers.tf
```diff
@@ -12,9 +12,14 @@ terraform {
       source  = "hashicorp/azurerm"
       version = "<= 3.99.0"
     }
+
+    restapi = {
```
gunzip, line 16: ditto

infra/_modules/networking/main.tf
```diff
@@ -4,8 +4,8 @@
 
 module "snet" {
   source               = "github.com/pagopa/terraform-azurerm-v3//subnet?ref=v8.3.0"
-  name                 = "${var.project}-${var.application_basename}-srch-snet-001"
-  resource_group_name  = var.resource_group_name
+  name                 = "${var.project}-${var.application_basename}-srch-snet-01"
+  resource_group_name  = "${var.project}-common-rg-01"
```
Krusty93, line 8: I'd restore the previous value where the resource group name is given by the caller, which has the actual reference

christian-calabrese: It has been changed cause the creation of the snet would fail if the resource group differs from the one of the vnet:
image

Krusty93 (approved):

rodoviti (approved):

PR #817, opened 2024-04-24
[IOPAE-1122] ADD Get Service By Id Azure Function

giamma1295:
ADD Get Service By Id Azure Function `GET /services/{serviceId}`
#### List of Changes
- ADD `GET /services/{serviceId}` HttpTrigger AZF
- ADD CosmosDb Dependecy
- ADD related Unit tests

Unit tests

#### Checklist:


apps/app-backend/src/functions/get-service-by-id.ts
```diff
@@ -0,0 +1,95 @@
...
+          )
+      ),
+      TE.chain((rr) =>
+        pipe(
+          rr,
+          O.fromPredicate(() => rr.statusCode === 200),
+          O.fold(
+            () =>
+              E.left(
+                new H.HttpNotFoundError(`Service '${serviceId}' not found`)
```
rodoviti, line 48: **question (blocking)**: are you sure this case happens only for not found item?

I mean, the response status code can be different from either 200 or 404 (rarely, but it can happen).

giamma1295: 598a7c503c5391f309586d5b3cdc5cf51cd243ab and 808189fd3f654572d64a287b3da1970fe7f21952

rodoviti (approved):
LGTM

PR #821, opened 2024-05-02
[IOPAE-1121] Create Service Details Read Model

StefanoDoc:
Create two cosmosdbTriggers that read from `services-publication` and `services-lifecycle` respectively to create a read model (new container services in a new db services-app) containing the service data necessary for viewing the service tab.

#### List of Changes

- Add new `request-detail` queue and config
- Add `service-details` container config
- Add 2 new watchers on service lifecycle and publication
- Add new `RequestDetailItem` model for queue and `ServiceDetails` model
- Add `onRequestDetail` function and `requestDetail` handler
- Unit tests

#### Checklist:


apps/io-services-cms-webapp/src/watchers/on-service-detail-lifecycle-change.ts
```diff
@@ -0,0 +1,43 @@
...
+      return pipe(item, onDetailLifecycleHandler, TE.right);
+    case "approved":
+    case "deleted":
+    case "rejected":
+    case "submitted":
+      return TE.right(noAction);
+    default:
+      return TE.right(noAction);
+  }
+};
```
giamma1295, line 43: You can simplify this switch in this way:
``` lang=ts
export const handler: RTE.ReaderTaskEither<
  { item: ServiceLifecycle.ItemType },
  Error,
  NoAction | OnDetailLifecycleActions
> = ({ item }) => pipe(
    item.fsm.state === "draft",
    B.fold(
        () => noAction,
        () => pipe(item, onDetailLifecycleHandler),
    ),
   TE.right
);
```

StefanoDoc: [4a69454](https://github.com/pagopa/io-services-cms/pull/821/commits/4a6945420a808640b31b8c13f42d1a7309c8520c)

giamma1295 (approved):
LGTM

PR #845, opened 2024-05-15
[EC-288] Set scheduling of indexers with 1 hour interval

christian-calabrese:
#### List of Changes
Added datachange policy to the organizations datasource to allow incremental indexing
Set scheduling of the three indexers with a 1 hour interval

#### Motivation and Context
The services and organizations indexes must be updated automatically to follow the addition and/or deletion of records in cosmosdb

#### Checklist:


giamma1295 (approved):
LGTM

PR #852, opened 2024-05-23
[IOPAE-1184] App backend health checks

giamma1295:
Add App backend health checks
#### List of Changes
- Add `CosmosDB` Health Check
- Add `AzureSearch` Health Check
- Add `AzureStorage` Health Check

#### Checklist:

### DIFF
--- a/apps/app-backend/src/utils/azure-search/health-check.ts
+++ b/apps/app-backend/src/utils/azure-search/health-check.ts
@@ -1,0 +1,20 @@
+import { pipe } from "fp-ts/lib/function";
+
+import * as TE from "fp-ts/lib/TaskEither";
+
+import {
+  HealthCheck,
+  toHealthProblems,
+} from "@pagopa/io-functions-commons/dist/src/utils/healthcheck";
+import { AzureSearchClientDependency } from "./dependency";
+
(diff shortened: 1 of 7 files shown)


apps/app-backend/src/utils/azure-search/health-check.ts
```diff
@@ -0,0 +1,20 @@
...
+import { AzureSearchClientDependency } from "./dependency";
+
+export type AzureSearchProblemSource = "AzureSearch";
+
+export const makeAzureSearchHealthCheck = <T>({
+  searchClient,
+}: AzureSearchClientDependency<T>): HealthCheck<AzureSearchProblemSource> =>
+  pipe(
+    searchClient.fullTextSearch({ searchText: "test" }),
+    TE.mapLeft(toHealthProblems("AzureSearch" as const)),
```
gunzip, line 18: **nit**: not needed for this PR but generally it's better to call conversion methods with the prefix "from" (instead that "to"). why? because in this way you can have multiple `fromOriginalType` methods (ie. `fromX(x: X)` `fromY(y: Y)`) **in the same module**: the module of the target type; while you can have _only one_ `toX` method in the module.

StefanoDoc (approved):

PR #859, opened 2024-05-27
[EC-311] change AI search sku to standard

christian-calabrese:
#### List of Changes

Created sku variable in AI Search module
Changed AI Search sku to standard

⚠️ Apply will be executed locally to overcome issues caused by the restapi provider that make the plan fail.
#### Motivation and Context

The current sku (basic) does not allow to scale above 3 replicas. This capacity may not be enough for production loads, hence the standard sku has been chosen.

#### Checklist:

### DIFF
--- a/infra/_modules/ai_search/variables.tf
+++ b/infra/_modules/ai_search/variables.tf
@@ -106,3 +106,14 @@ variable "indexers_scheduling_interval" {
     services_publication = "PT1H"
   }
 }
+
+variable "sku" {
+  type        = string
+  description = "The SKU of the Search Service"
+  default     = "standard"
+
+  validation {
+    condition     = contains(["basic", "free", "standard", "standard2", "standard3", "storage_optimized_l1", "storage_optimized_l2"], var.sku)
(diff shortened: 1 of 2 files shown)


rodoviti (approved):

infra/_modules/ai_search/variables.tf
```diff
@@ -105,4 +105,15 @@ variable "indexers_scheduling_interval" {
     services_lifecycle   = "PT1H"
     services_publication = "PT1H"
   }
+}
+
+variable "sku" {
+  type        = string
+  description = "The SKU of the Search Service"
+  default     = "standard"
```
giamma1295, line 113: May it be "StandardS1" the sku name?

christian-calabrese: Initially I thought the same. However, from the [resource documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/search_service#arguments-reference),  it looks like "standard" is the right one

giamma1295 (approved):
LGTM

PR #867, opened 2024-05-31
[IOPAE-1201] Remove unwanted continuationToken on serviceHistory first call

giamma1295:
Currently the first call to the ServiceHistory API includes an unintended continuationToken(react passes a SyntheticBaseEvent which is mistakenly then used in the call as a continuationToken), as shown in the following screen.

image

#### List of Changes
- remove any Parameter on serviceHistory first call

Local Application Run

#### Checklist:


giuseppedipinto (approved):
LGTM

PR #879, opened 2024-06-04
Use new azure storage SDK on io-services-app-backend

giamma1295:
Use `@azure/storage-blob` instead of the old `azure-storage`
Unit Tests

#### Checklist:


StefanoDoc (approved):

rodoviti (approved):
LGTM

apps/app-backend/src/functions/featured-services.ts
```diff
@@ -1,50 +1,43 @@
...
-        () =>
-          getBlobAsObject(
-            FeaturedServices,
-            blobService,
-            featuredItemsConfig.FEATURED_ITEMS_CONTAINER_NAME,
-            featuredItemsConfig.FEATURED_SERVICES_FILE_NAME
-          ),
-        (err) =>
-          new H.HttpError(`Unexpected error: [${E.toError(err).message}]`)
+      getBlobAsObject(
```
gunzip, line 28: what about chaing `getBlobAsObject` into a RTE itself?

PR #880, opened 2024-06-04
Temporary storage account for static-content imported

christian-calabrese:
#### List of Changes

Defined and imported temporary storage account for static-content files
Added private endpoints to the storage account (one for each subservice to make app healthchecks work properly)
Added the storage account connection string to the function app environment variables

#### Motivation and Context

The new [function app module](https://github.com/pagopa/dx/tree/main/infra/modules/azure_function_app) forces the usage of managed identities when authenticating to the function app's storage account. However, the library used in the backend does not support managed identities.

For this reason a new temporary storage account has been created via azure portal and imported with this PR. Additional resources as private endpoints and environment variable are added as well.

Terraform plan

#### Checklist:


giamma1295 (approved):

christian-calabrese:
The temporary storage account is not needed anymore. The function now works correctly with the managed identity and the internal storage account.

PR #889, opened 2024-06-05
[#IOPAE-1198]Add privacy_url and contact fields to automatic controls

StefanoDoc:
According to documentation , privacy_url and one of the 4 contact fields are mandatory for a service to be published , so we need to check if is set when the service is ready for review , in addition if the service is in the `SERVICEID_QUALITY_CHECK_EXCLUSION_LIST` , then we don't need this controls
#### List of Changes

add `ServiceQualityStrict` type
add `isServiceAllowedForQualitySkip`
update `createServiceValidationHandler` and `validate` function
unit test

#### Checklist:

### DIFF
--- a/.changeset/lovely-hornets-whisper.md
+++ b/.changeset/lovely-hornets-whisper.md
@@ -1,0 +1,6 @@
+---
+"@io-services-cms/models": patch
+"io-services-cms-webapp": patch
+---
+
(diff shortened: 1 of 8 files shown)


rodoviti (approved):
LGTM

giamma1295 (approved):

PR #907, opened 2024-06-13
[IOPAE-1181] IP inputs unclickable

carlo-bussi:
#### List of Changes

Now the IP inputs are unclickable outiside the edit mode

#### Motivation and Context

QA

Local env

https://github.com/pagopa/io-services-cms/assets/157382565/5232c160-4b80-4eba-906a-be0802396e1a

#### Checklist:

### DIFF
--- a/apps/backoffice/src/components/forms/controllers/text-field-array-controller.tsx
+++ b/apps/backoffice/src/components/forms/controllers/text-field-array-controller.tsx
@@ -194,7 +194,8 @@ export function TextFieldArrayController({
               variant="contained"
               onClick={() => {
                 onSaveClick && onSaveClick();
-                setIsReadOnly(true);
+                setRenderedFields({ ...renderedFields });
+                readOnly && setIsReadOnly(true);
               }}
             >
               {t("buttons.save")}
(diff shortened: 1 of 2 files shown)


apps/backoffice/src/components/forms/controllers/text-field-array-controller.tsx
```diff
@@ -119,6 +122,13 @@ export function TextFieldArrayController({
                     {...register(getFieldName(index))}
                     {...props}
                     margin="normal"
+                    sx={
```
gunzip, line 125: what about adding a `disabled` attribute?

carlo-bussi: we want to keep the standard input style. add disable and change its style would be more complicated

giuseppedipinto (commented):
Please check this component in service create/update mode: in both cases we have a regression in the behaviour because we are missing out the edit functionality as there is no possibility to add or edit a new or existent IP field.

giuseppedipinto (changes requested):
It still doesn't work, steps to reproduce the problem:

1. Try creating a service
2. add an ip/cidr field
3. add a second ip/cidr field
4. click _Cancel_ button
5. add an ip/cidr field
6.❌ Fail edit

- Same behaviour in service edit use case.
- Same behaviour _(maybe even worse)_ in Api Key editing.

giuseppedipinto (approved):
LGTM

PR #935, opened 2024-07-02
chore:[IOPAE-1073] Fix last update version

giamma1295:
Fix last_update and version on cosmos collections.

#### List of Changes
- Avoid saving `last_update` and `version`, which are API response representations for `_ts` and `_etag` cosmos record fields, in `services-lifecycle` and `service-publication` cosmos containers.
- Fix the way how `last_update` and `version` are valued in `service-history` cosmos containers, now they are based on `services-lifecycle` and `service-publication` `_ts` and `_etag` fields.

Unit tests

#### Checklist:


apps/io-services-cms-webapp/src/watchers/on-service-publication-change.ts
```diff
@@ -16,18 +16,22 @@ type OnReleaseActions = RequestHistoricizationAction;
...
+}: ServicePublication.CosmosResource): RequestHistoricizationAction => ({
   requestHistoricization: {
-    ...item,
-    last_update:
-      item.last_update ?? (new Date().toISOString() as NonEmptyString), // last_update fallback (value is always set by persistence layer) TODO add log
+    ...otherProps,
+    // eslint-disable-next-line no-underscore-dangle
+    last_update: new Date(_ts * 1000).toISOString() as NonEmptyString, // last_update on service-history record corresponds to the _ts on the service-publication record
+    // eslint-disable-next-line no-underscore-dangle
+    version: _etag as NonEmptyString, // version on service-history record corresponds to the _etag on the service-publication record
```
rodoviti, line 29: **suggestion (non-blocking):** avoid code duplication.

These lines are the same as [on-service-lifecycle-change.ts#onAnyChangesHandler](https://github.com/pagopa/io-services-cms/pull/935/files#diff-31e3e4997d71ad64bb849e0f1ad117d48e12b463d9408a30f93b150a2e975d76R44)...why not centralize this logic in [apps/io-services-cms-webapp/src/historicizer/request-historicization-handler.ts#toServiceHistory](https://github.com/pagopa/io-services-cms/pull/935/files#diff-8d20c959284cf903aa267213021b75fbc82b715b6ea1f2eb29ae7150b14bd689R24)?

apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
```diff
@@ -35,29 +35,34 @@ type RequestHistoricizationAction = Action<
...
 const onSubmitHandler = (
-  item: ServiceLifecycle.ItemType
+  item: ServiceLifecycle.CosmosResource
 ): RequestReviewAction => ({
   requestReview: {
     id: item.id,
     data: item.data,
-    version: item.version ?? (`ERR_${ulidGenerator()}` as NonEmptyString), // TODO add log
+    // eslint-disable-next-line no-underscore-dangle
+    version: item._etag ?? (`ERR_${ulidGenerator()}` as NonEmptyString), // TODO add log
```
rodoviti, line 59: **chore (non-blocking):** remove useless logic.

Now, using `_etag` field, this logic is no more useful (it will always be valued)

cc: @giuseppedipinto I'm tagging you for a double check, since you added this logic at the time 😉

apps/io-services-cms-webapp/src/watchers/on-service-publication-change.ts
giamma1295: e7143dee0696d567cb39874ff987014daf0a275f

apps/io-services-cms-webapp/src/watchers/on-service-lifecycle-change.ts
giamma1295: e7143dee0696d567cb39874ff987014daf0a275f

rodoviti (approved):
LGTM

PR #945, opened 2024-07-12
[#IOPAE-1305] add missing Backoffice build step to code review

StefanoDoc:
#### List of Changes

- update turbo to `^2.0.7`
- fix turbo repo config
- remove generate step from code-review because now turbo repo takes care of it
- remove generate step from BO build because it causes a double generation now that turbo takes care of it
- add backoffice build step

local test and [test pr](https://github.com/pagopa/io-services-cms/pull/944)

Screenshot 2024-07-11 alle 15 10 13

#### Checklist:

### DIFF
--- a/apps/backoffice/package.json
+++ b/apps/backoffice/package.json
@@ -15,6 +15,7 @@
     "coverage": "yarn coverage:b4f && yarn coverage:fe",
     "coverage:b4f": "vitest run --config ./vitest.config.node.ts --coverage",
     "coverage:fe": "vitest run --config ./vitest.config.react.ts --coverage",
+    "typecheck": "next build",
     "api-definitions:clean": "rm -rf ./src/generated/api",
     "api-definitions:generate": "gen-api-models --api-spec ./openapi.yaml --out-dir ./src/generated/api --request-types --response-decoders --client",
     "api-definitions:all": "yarn api-definitions:clean && yarn api-definitions:generate",
--- a/apps/backoffice/package.json
+++ b/apps/backoffice/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "yarn generate && next build",
+    "build": "next build",
     "build:pre-package": "cp -r .next/static .next/standalone/apps/backoffice/.next && cp -r public .next/standalone/apps/backoffice",
     "build:package": "yarn build:pre-package && (cd .next/standalone/apps/backoffice && zip -r -q ../../../../${npm_package_name}-${npm_package_version}.zip . -x package.json)",
     "start": "./node_modules/next/dist/bin/next start",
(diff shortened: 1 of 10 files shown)


apps/backoffice/package.json
```diff
@@ -15,6 +15,7 @@
     "coverage": "yarn coverage:b4f && yarn coverage:fe",
     "coverage:b4f": "vitest run --config ./vitest.config.node.ts --coverage",
     "coverage:fe": "vitest run --config ./vitest.config.react.ts --coverage",
+    "typecheck": "next build",
```
rodoviti, line 18: In turboRepo config, `typecheck` step dependsOn `build` step, so i think with this config next app will be built two times....is there a way to add a typecheck step that "do nothing"?

rodoviti: **question:** Have you tried with `"typecheck": "tsc --noEmit"` instead of `"typecheck": "next build"`?

StefanoDoc: 1) i don't know if you can set a step that does nothing (i will try and see) , but seeing the code-review log it does not seem like it gets done 2 times , plus the code-review does not call the build step , but maybe it's not connected ? , sorry i don't know a lot about turbo

2) i tried and it doesn't work as you can see in the "[test pr](https://github.com/pagopa/io-services-cms/pull/944)" , because of this inside a node-module(next/image-types)
`declare module '*.svg' {
  /**
   * Use `any` to avoid conflicts with
   * `@svgr/webpack` plugin or
   * `babel-plugin-inline-react-svg` plugin.
   */
  const content: any

  export default content
}`
that is needed for the .svg files

StefanoDoc: i tried a "" for typecheck and i simply go back to the previous version and the as-is

StefanoDoc: from what i see the build step is being done automatically only for whats inside `packages` and not whats inside `apps` even though the config should be the same , so i'm pretty sure the build is not being done twice but i don't know why this is happening , because in theory it should be as you say i think
Screenshot 2024-07-12 alle 16 09 58

turbo.json
```diff
@@ -1,13 +1,7 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "pipeline": {
-    "generate": {
-      "inputs": [
-        "./openapi.yaml",
-        "src/generated/**/*"
-      ],
-      "outputMode": "errors-only"
```
rodoviti, line 9: **chore:** add `"outputMode": "errors-only"` 😉

rodoviti (changes requested):

turbo.json
StefanoDoc: [80d4cf1](https://github.com/pagopa/io-services-cms/pull/945/commits/80d4cf1ea4efdfa097bb2368c7d9d6315be9abad)

rodoviti (approved):
LGTM

rodoviti (approved):

christian-calabrese (approved):

PR #946, opened 2024-07-16
[IOPAE - 1287] Header component tests

carlo-bussi:
#### List of Changes

Added tests for Header component

#### Motivation and Context

Testing
local env

#### Checklist:


apps/backoffice/src/components/headers/__tests__/page-html-head-title..test.tsx
```diff
@@ -0,0 +1,37 @@
+import { cleanup, render, screen } from "@testing-library/react";
+import React from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { PageHtmlHeadTitle } from "../page-html-head-title";
+
+let section = "";
```
giuseppedipinto, line 6: what do you think if we added a "test" section in order to verify in the expected that this value is present after the "|" of line 33?

giuseppedipinto (changes requested):
1. It might be useful to add unit tests also for `apps/backoffice/src/components/headers/page-header.tsx` ?
2. What do you think if we used this PR to move the `apps/backoffice/src/components/__tests__/page-breadcrumbs.test.tsx` file into the new `apps/backoffice/src/components/headers/__tests__` folder?

giuseppedipinto (approved):

PR #960, opened 2024-07-19
[EC-369] Use AD authentication instead of connection string for TF state

Krusty93:
#### List of Changes

Tell Terraform to use AD authentication rather to connect to the Storage Account which hold the state

#### Motivation and Context

The connection string authentication method is disabled on a Storage Account

#### Checklist:

### DIFF
--- a/.identity/env/prod/terraform.tfvars
+++ b/.identity/env/prod/terraform.tfvars
@@ -71,6 +71,9 @@ infra_environment_cd_roles = {
     ],
     io-p-services-cms-rg = [
       "Role Based Access Control Administrator"
+    ],
+    io-p-services-cms-rg = [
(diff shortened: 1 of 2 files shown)


giamma1295 (approved):

christian-calabrese (approved):

.identity/env/prod/terraform.tfvars
```diff
@@ -69,6 +69,9 @@ infra_environment_cd_roles = {
       "Storage Blob Data Contributor",
       "Storage Queue Data Contributor",
     ],
+    io-p-services-cms-rg = [
+      "Role Based Access Control Administrator"
+    ],
```
rodoviti, line 74: @Krusty93 these lines are already there!

PR #983, opened 2024-08-22
[IOPAE-931] Handle Constraint violation on insert, map permanent errors

giamma1295:
Error Handling in QueueTriggers.
#### List of Changes
- Handle `constraint violation` on ticket reference insert in `service-review-checker`.
- Map error on incoming message decoding as permanent errors in order to avoid retry in QueueTrigger Functions.

Unit Test

#### Checklist:


apps/io-services-cms-webapp/src/reviewer/request-review-handler.ts
```diff
@@ -88,7 +95,18 @@ const saveTicketReference =
         status: "PENDING",
         extra_data: {},
       }),
-      TE.mapLeft(E.toError)
+      TE.map((_) => void 0),
+      TE.orElseW((e) =>
+        pipe(
+          // Check if the error is due to a unique constraint violation
+          // If so, we can ignore it, cause it means that the ticket reference is already present in the db
+          !!e.constraint,
```
iwoak, line 103: This field also exists for other types of error, although this application cannot generate them now (such as a foreign key violation), it may be safer to use the code field, what do you think?

https://www.postgresql.org/docs/current/errcodes-appendix.html
23505 is the primary key violation error
23503 is the foreign key violation error

giamma1295: mhh, yes we tested and we can handle this Kind of error also by intercepting the PK violation code 23505!

giamma1295: done in [35219da](https://github.com/pagopa/io-services-cms/pull/983/commits/35219dad3e0f8c756af755e00d8caea804f3ee9b)

StefanoDoc (approved):

PR #990, opened 2024-08-30
[IOPAE-1349]custom update dt field

giamma1295:
Add new `modified_at` field on CosmosDB `service-lifecycle` and `service-publication` containers , this will allow to better manage service update date and to backport the original service update date for migrated services.

#### List of Changes
- Add `modified_at` field in `service-lifecycle` and `service-publication` container records.
- Edit `store.cosmos` in order to handle the new `modified_at` field:
  - on `save method` will fill the field with the current Timestamp before write the record on CosmosDB
  - on `fetch and bulkFetch methods` will lazy init this field with the content in `_ts`(converted from seconds to millis), this behaviour is needed to be fault tollerant, previously wrote records does not have this field(we will fix up those records using a fix script later on).
- Update All `CosmosDBTrigger` and `QueueTrigger` who was using `last_update` to use the new `modified_at` field
- Update All converters(CosmosDB Record Types <-> Api Payload Response Types) in order to use the new field.
- Remove unnecessary `last_update` field in CosmosDB Record Type, this field is a runtime calculated field erroneusly present in those types.
- Updated tests.

#### Motivation and Context

This change will allow us to resolve the misleading situation of service update dates between the `service-publication/service-lifecycle` containers and the `service-history` container, especially for migrated services that have:
- in `service-publication/service-lifecycle` as the update date the date they were migrated, since the `_ts` field of cosmosDb is used.
- in `service-history`, which uses custom fields, `last_update` (ISO UTC string) and `id` (timestamp with UTC time zone), the actual update date reported by the legacy `services` container.

Local Application run(only HttpTriggers) and Unit Tests.

#### Checklist:


apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
```diff
@@ -89,6 +89,7 @@ const fromLegacyToCmsService = (
     status === "published" || status === "unpublished"
       ? "PublicationItemType"
       : "LifecycleItemType",
+  last_update_ts: Math.floor(Date.now() / 1000),
```
rodoviti, line 92: **question:** if the change comes from the Legacy application, it shouldn't be `last_update_ts: service._ts`?

apps/io-services-cms-webapp/src/synchronizer/request-sync-cms-handler.ts
```diff
@@ -113,12 +113,6 @@ export const handleQueueItem = (
   pipe(
     queueItem,
     parseIncomingMessage(Queue.RequestSyncCmsItems),
-    E.mapLeft(
-      (err) =>
-        new Error(
-          `Error while parsing incoming message, the reason was => ${err.message}`,
-        ),
-    ), // TODO: map as _permanent_ error
```
rodoviti, line 121: **question:** Why are these lines no longer useful?

apps/io-services-cms-webapp/src/historicizer/request-historicization-handler.ts
```diff
@@ -44,7 +47,6 @@ export const handleQueueItem = (context: Context, queueItem: Json) =>
     queueItem,
     parseIncomingMessage(Queue.RequestHistoricizationItem),
     TE.fromEither,
-    TE.mapLeft((_) => new Error("Error while parsing incoming message")), // TODO: map as _permanent_ error
```
rodoviti, line 47: **question:** Why is this line no longer useful?

apps/io-services-cms-webapp/src/synchronizer/request-sync-cms-handler.ts
giamma1295: Last week we did a little refactor on queue handlers, the parseIncomingMessage was replicated in all the handler so we made a generic version, that you can find in io-services-cms/apps/io-services-cms-webapp/src/utils/queue-utils.ts, this mapLeft is a leftover from this refactoring

rodoviti (commented):
question: Why persist the Unix Timestamp in seconds and not milliseconds?

Millisecond Unix Timestamp is more accurate and is the default timestamp used by, in our case, Node.js; therefore, this format saves us from performing a “conversion” every time we need to return it in its textual representation (`new Date(last_update_ts * 1000).toISOString()`). If the choice of Unix Timestamp seconds is only for compatibility with CosmosDB's _ts field, then I think perhaps it is convenient to use the millisecond format, limiting the conversion from seconds to milliseconds only in (limited, I think) cases where this new field is not yet valued. Don't you think?

EDIT:

> on save method will fill the field with the current Timestamp, seconds @ UTC zone(this format is used on CosmosDB for the _ts, and the id field in service-history) before write the record on CosmosDB

There are already items in the `service-history` container with `id = _ts` (seconds Unix Timestamp), so unless you recreate that container (resetting its feeds), it is not possible.

apps/io-services-cms-webapp/src/historicizer/request-historicization-handler.ts
giamma1295: same as the previous comment!

rodoviti: 👌  replied in this [comment](https://github.com/pagopa/io-services-cms/pull/990#discussion_r1743350955)

rodoviti:
**suggestion (non-blocking):** change the new field name.

What do you think to change the new field from `last_update_ts` to `modified_at` (or something like that)?
This way the name is “unique,” especially with respect to the other (computed) field `last_update`, and avoids entering the “type” name (_ts_) as a suffix.

packages/io-services-cms-models/src/lib/fsm/store.cosmos.ts
```diff
@@ -115,7 +111,12 @@ export const createCosmosStore = <
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       ({ last_update, version, ...valueToSave }) =>
         TE.tryCatch(
-          () => container.items.upsert({ ...valueToSave, id }),
+          () =>
+            container.items.upsert({
+              ...valueToSave,
+              id,
+              last_update_ts: Math.floor(Date.now() / 1000),
```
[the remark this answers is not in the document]
giamma1295: done!

apps/io-services-cms-webapp/src/watchers/on-legacy-service-change.ts
giamma1295: done!

apps/io-services-cms-webapp/src/utils/converters/service-lifecycle-converters.ts
```diff
@@ -76,7 +76,9 @@ export const itemToResponse =
         (err) => ResponseErrorInternal(err.message),
         (topic) => ({
           id,
-          last_update: last_update ?? new Date().getTime().toString(),
+          last_update: modified_at
+            ? DateUtils.isoStringfromUnixTimestamp(modified_at)
+            : new Date().getTime().toString(),
```
rodoviti, line 81: chore: fix `toISOString()` instead of `toString()`
```suggestion
            : new Date().getTime().toISOString(),
```

giamma1295: Should be this `new Date().toISOString()`

giuseppedipinto:
> question: Why persist the Unix Timestamp in seconds and not milliseconds?
> 
> Millisecond Unix Timestamp is more accurate and is the default timestamp used by, in our case, Node.js; therefore, this format saves us from performing a “conversion” every time we need to return it in its textual representation (`new Date(last_update_ts * 1000).toISOString()`). If the choice of Unix Timestamp seconds is only for compatibility with CosmosDB's _ts field, then I think perhaps it is convenient to use the millisecond format, limiting the conversion from seconds to milliseconds only in (limited, I think) cases where this new field is not yet valued. Don't you think?
> 
> EDIT:
> 
> > on save method will fill the field with the current Timestamp, seconds @ UTC zone(this format is used on CosmosDB for the _ts, and the id field in service-history) before write the record on CosmosDB
> 
> There are already items in the `service-history` container with `id = _ts` (seconds Unix Timestamp), so unless you recreate that container (resetting its feeds), it is not possible.

In general I would also agree with setting the date in milliseconds.
For `service-history` we would do the milliseconds/seconds adaptation or, in case of data recreation, there should be no problem.

packages/io-services-cms-models/src/utils/date-utils.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * Returns the ISOString rapresentation for a given UNIX timestamp.
+ *
+ * @param {Number} unixTimestamp - UNIX timestamp to be converted to ISOString
+ * @returns {String}
+ */
+const isoStringfromUnixTimestamp = (unixTimestamp: number): string =>
```
rodoviti, line 7: **suggestion (non-blocking):** rename function or change description.

This function could be misunderstood and in some cases not work properly. The function says to translate the timestamp passed as input into ISO String, but in reality it currently translates it correctly only if the timestamp is expressed in milliseconds. If you pass it a timestamp expressed in seconds the translation will not be correct!
I suggest slightly modifying the name of fn to make it clearer, or better describe its behavior in the documentation.

packages/io-services-cms-models/src/utils/date-utils.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @returns {Number}
+ */
+const unixSecondsToMillis = (unixSeconds: number): number => unixSeconds * 1000;
+
+/**
+ * Returns the UNIX timestamp for the given `date`. Defaults to `new Date()`
+ * when not providing the `date` argument to the method call.
+ *
+ * @param {Date} date - The date to convert to UNIX timestamp
+ * @returns {Number}
```
rodoviti, line 41: **suggestion (non-blocking):** make function return "type" clearer
```suggestion
 * @returns {Number} the timestamp expressed in milliseconds
```

packages/io-services-cms-models/src/lib/fsm/store.cosmos.ts
```diff
@@ -115,7 +119,12 @@ export const createCosmosStore = <
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       ({ last_update, version, ...valueToSave }) =>
         TE.tryCatch(
-          () => container.items.upsert({ ...valueToSave, id }),
+          () =>
+            container.items.upsert({
+              ...valueToSave,
+              id,
+              modified_at: unixTimestamp(),
```
rodoviti, line 126: **issue:** there is a case for which the `modified_at` field should not be overwritten with current timestamp!

When synching from legacy, the field modified_at will be filled with legacy timestamp and that timestamp should be saved as `modified_at`, not the current one!

apps/io-services-cms-webapp/host.json
```diff
@@ -38,5 +38,17 @@
...
+    "GetServiceLifecycle",
+    "GetServicePublication",
+    "GetServiceHistory",
+    "CreateService",
+    "EditService",
+    "DeleteService",
+    "PublishService",
+    "UnPublishService",
+    "ReviewService"
+  ]
```
rodoviti, line 53: chore: I think this is an oversight 😉 
```suggestion
  }
```

giamma1295: yup, pushed by mistake, removed

giamma1295: a0482a604321c423e215cc545208d89e2095079d

packages/io-services-cms-models/src/lib/fsm/store.cosmos.ts
giamma1295: 93edf2f6f8de40684f88f812d787b6a44a74a775

packages/io-services-cms-models/src/utils/date-utils.ts
giamma1295: 93edf2f6f8de40684f88f812d787b6a44a74a775

rodoviti (approved):
LGTM

PR #995, opened 2024-09-03
refactor:remove getUserByEmail calls

StefanoDoc:
Optimization change for GetServices and CreateServices first one by removing a call to APIM second one using a method that uses an `Id` to search instead of an `email`
#### List of Changes
GetServices -> avoid an APIM call to retrieve the ApimUserId....this info is already available from x-user-id header
CreateServices -> using GetUsers instead of GetUserByEmail should speed up the api

local test

Screenshot 2024-09-03 alle 17 10 40
Screenshot 2024-09-03 alle 17 11 23

Screenshot 2024-09-03 alle 17 00 00

#### Checklist:

### DIFF
--- a/.changeset/brown-dolls-march.md
+++ b/.changeset/brown-dolls-march.md
@@ -1,0 +1,5 @@
+---
+"io-services-cms-webapp": patch
+---
+
+remove getUserEmail calls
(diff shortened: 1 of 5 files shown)


rodoviti (approved):
LGTM

PR #997, opened 2024-09-03
[IOPAE-1343] set new cms headers

rodoviti:
Needed for the management of selfcare groups

#### List of Changes

- [set new CMS headers needed for handling Selfcare Groups](https://github.com/pagopa/io-services-cms/commit/1873179b5b881e10297ab6b8acdad259769d47fe)

Unit Tests

#### Checklist:


giamma1295 (approved):

infra/src/api/io_services_cms/v1/policy.xml
```diff
@@ -20,6 +20,12 @@
         <set-header name="x-user-email" exists-action="override">
             <value>@(context.User.Email)</value>
         </set-header>
+        <set-header name="x-user-groups-selc" exists-action="override">
```
Krusty93, line 23: I propose

```xml
<choose>
    <when condition="@(context.Subscription.Id.StartsWith("MANAGE-GROUP-")">
        <set-header name="x-user-groups-selc" exists-action="override">
            <value>@(context.Subscription.Id.Substring("MANAGE-GROUP-".Length))</value>
        </set-header>
   </when>
</choose>
```

Krusty93 (approved):

infra/src/api/io_services_cms/v1/policy.xml
rodoviti: In this case, if the client sets this header (externally) and the condition is not met, the header valued by the client would be propagated to the backend, which we want to avoid since this header is meant to be handled only “internally”

PR #1004, opened 2024-09-10
[IOPAE-1372] fix regenerate service keys on deleted services

StefanoDoc:
#### Checklist:


apps/io-services-cms-webapp/src/webservice/controllers/regenerate-service-keys.ts
```diff
@@ -55,12 +58,35 @@ type RegenerateServiceKeysHandler = (
...
+          O.map(() => void 0),
+          TE.fromOption(() =>
+            ResponseErrorNotFound(
+              "Not found",
+              `no item with id ${serviceId} found`,
+            ),
+          ),
+        ),
+      ),
+    );
```
giamma1295, line 84: This function can be moved to a different module and imported both in here and in `get-service-key` controller!

giamma1295 (approved):

PR #1009, opened 2024-09-12
fix path for clean and generation selfcare-api-models in backoffice

giamma1295:
Fix path in `clean:selfcare-api-models` yarn script and `generate:selfcare-api-models` turbo task.
#### Checklist:


rodoviti (approved):
LGTM

PR #1011, opened 2024-09-16
[iopae-1382] avoid setting groupid via apim

rodoviti:
Avoid setting groupid via apim

#### List of Changes

- change createService policy to handle new groupId field (hidden from user)
- add updateService policy to handle new groupId field (hidden from user)

Tested through the APIM "echo api"

#### Checklist:

### DIFF
--- a/infra/src/README.md
+++ b/infra/src/README.md
@@ -37,6 +37,7 @@
 | [azurerm_api_management_api_diagnostic.services_cms_api_app_insights](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_api_diagnostic) | resource |
 | [azurerm_api_management_api_operation_policy.create_service_policy_v2](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_api_operation_policy) | resource |
 | [azurerm_api_management_api_operation_policy.get_service_topics_policy_v2](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_api_operation_policy) | resource |
+| [azurerm_api_management_api_operation_policy.update_service_policy_v2](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_api_operation_policy) | resource |
 | [azurerm_api_management_logger.cache_policy_app_insights](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_logger) | resource |
 | [azurerm_cosmosdb_sql_database.db_app_be](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cosmosdb_sql_database) | resource |
 | [azurerm_cosmosdb_sql_database.db_cms](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cosmosdb_sql_database) | resource |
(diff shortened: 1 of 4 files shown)


infra/src/api/io_services_cms/v1/createservice_policy/policy.xml
```diff
@@ -6,6 +6,30 @@
...
+    var jsonBody = JObject.Parse(stringBody);
+    try {
+      if (context.Request.Headers.ContainsKey("x-user-groups-selc")) {
+        jsonBody["data"]["groupId"] = context.Request.Headers["x-user-groups-selc"][0];
+      } else {
+        jsonBody["data"]["groupId"].Parent.Remove();
+      }
+      return jsonBody.ToString();
+    } catch (Exception) {
+      return jsonBody.ToString();
```
Krusty93, line 24: Do you think it could be useful to use the `trace` policy to track when the exception is raised for troubleshooting purposes? https://learn.microsoft.com/en-us/azure/api-management/trace-policy

infra/src/api/io_services_cms/v1/createservice_policy/policy.xml
```diff
@@ -6,6 +6,30 @@
...
+                <set-body>
+@{
+  var stringBody = context.Request.Body.As<string>();
+  try {
+    var jsonBody = JObject.Parse(stringBody);
+    try {
+      if (context.Request.Headers.ContainsKey("x-user-groups-selc")) {
+        jsonBody["data"]["groupId"] = context.Request.Headers["x-user-groups-selc"][0];
+      } else {
+        jsonBody["data"]["groupId"].Parent.Remove();
```
Krusty93, line 20: Did you try this? I honestly don't remember if `Remove` is a void method or returns a reference to the existing collection without the removed data

infra/src/api/io_services_cms/v1/createservice_policy/policy.xml
rodoviti: Actually these policy are only useful only to ensure "robustness" of these two operations (createService and editService), because this new field (groupId) will not be added to "external" OpenAPI, so clients using these APIs should not even be aware of the existence of this field.
Actually, these exceptions could be raised only in these two cases:
- when the request body is not a json (`JObject.Parse(stringBody)` fails)
- when the request body has not a `data` property (`jsonBody["data"]["groupId"]` fails)
Both cases are managed by the application/backend request handler (and, in case of failure, also traces via Application Insight).

So, I think there is no need to add the trace here.

PS: an exception could be raised also in cases when caller do not insert groupId property (`jsonBody["data"]["groupId"].Parent.Remove()` fails), but, in this case, the absence of this field is exactly what we want, so it is OK to leave the request as it came to us.

rodoviti: Basically, all request checking/validation is done by the application/backend (and also traced via App Insight), this policy is only to ensure that from this channel (APIM) the groupId field is not manipulated by the caller.

infra/src/api/io_services_cms/v1/createservice_policy/policy.xml
rodoviti: I've tested both policy using the APIM "echo API" 😉

christian-calabrese (approved):

PR #1017, opened 2024-09-17
fix:[IOPAE-1373] GetInstitutionsGroups API

StefanoDoc:
#### List of Changes

- fix optional query param decode
- change `limit` and `offset` to `size` and `page` to match selfcare model
- add `PaginationResult` model in openapi
- fix mapping function

1. unit test
2. local backoffice start

#### Checklist:

### DIFF
--- a/apps/backoffice/src/app/api/institutions/[institutionId]/groups/route.ts
+++ b/apps/backoffice/src/app/api/institutions/[institutionId]/groups/route.ts
@@ -1,12 +1,12 @@
 import {
   isBackofficeUserAdmin,
-  isInstitutionIdSameAsCaller
+  isInstitutionIdSameAsCaller,
 } from "@/lib/be/authz";
 import {
   handleBadRequestErrorResponse,
   handleForbiddenErrorResponse,
   handleInternalErrorResponse,
-  handlerErrorLog
+  handlerErrorLog,
 } from "@/lib/be/errors";
 import { retrieveInstitutionGroups } from "@/lib/be/institutions/business";
 import { sanitizedNextResponseJson } from "@/lib/be/sanitize";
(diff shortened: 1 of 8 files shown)


giamma1295 (approved):
LGTM

giamma1295 (commented):
LGTM but add unit tests

apps/backoffice/src/app/api/institutions/[institutionId]/groups/route.ts
```diff
@@ -62,3 +52,14 @@ export const GET = withJWTAuthHandler(
...
+const getQueryParam = (
+  request: NextRequest,
+  param: string,
+  defaultValue: number,
+): number => {
+  const rawParam = NumberFromString.decode(
+    request.nextUrl.searchParams.get(param),
+  );
+  return E.isRight(rawParam) ? rawParam.right : defaultValue;
+};
```
Comment (rodoviti) [inline apps/backoffice/src/app/api/institutions/[institutionId]/groups/route.ts:65]: **chore:** improve validation

With the current implementation, if the queryParam is provided but it is not a valid number (i.e.: `size=foo`), it will ignore the error and continue with default values; but in that case we should return a Bad Request

Comment (giamma1295) [reply apps/backoffice/src/app/api/institutions/[institutionId]/groups/route.ts:65]: What do you think about a generic method to use when needed to retrieve and validate a queryParam?
Somenthin like this:

```typescript
const extractQueryParam = <A>(
  request: NextRequest,
  param: string,
  decoder: t.Type<A>,
  defaultValue?: A,
): E.Either<Error, A> =>
  pipe(
    request.nextUrl.searchParams.get(param),
    O.fromNullable,
    O.foldW(
      () =>
        defaultValue
          ? E.right(defaultValue)
          : E.left(new Error(`Missing '${param}' parameter in request`)),
      (rawParam) =>
        pipe(
          rawParam,
          decoder.decode,
          E.mapLeft((errors) => E.toError(readableReport(errors))),
        ),
    ),
  );

```

Comment (giamma1295) [reply apps/backoffice/src/app/api/institutions/[institutionId]/groups/route.ts:65]: We can also define an Error for bad request and intercept it into a try cath that wrap the controller.

Comment (giamma1295) [reply apps/backoffice/src/app/api/institutions/[institutionId]/groups/route.ts]: Ok! seems legit, we can create a dedicated task and address this later on!

rodoviti (approved):

PR #1018, opened 2024-09-18
[IOPAE-1383] MixPanel integration

carlo-bussi:
#### List of Changes

Integrated logging event to Mixpanel service.

#### Motivation and Context

We need to keep track of the user experience on the BackOffice

On local enviroment and Mixpanel Dashboard

#### Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.


apps/backoffice/src/utils/mix-panel.ts
```diff
@@ -0,0 +1,58 @@
+import mixpanel from "mixpanel-browser";
+
+type MixPanelEventsStructure = {
+  readonly IO_BO_LOGIN: {};
+  readonly IO_BO_OVERVIEW_PAGE: {};
+  readonly IO_BO_SERVICES_PAGE: {};
+  readonly IO_BO_SERVICE_DETAILS_PAGE: {
+    serviceId: string;
```
giuseppedipinto, line 8: choose and use only one notation: camelCase ora snake_case for obj properties

apps/backoffice/src/pages/services/[serviceId]/edit-service.tsx
```diff
@@ -51,6 +56,10 @@ export default function EditService() {
           title: t("notifications.validationError"),
         }),
       );
+      logToMixpanel("IO_BO_SERVICE_CREATE_END", {
```
Comment (giuseppedipinto) [inline apps/backoffice/src/pages/services/[serviceId]/edit-service.tsx:59]: Should be `IO_BO_SERVICE_EDIT_END`, correct?

apps/backoffice/src/pages/services/index.tsx
```diff
@@ -454,7 +469,14 @@ export default function Services() {
               passHref
               style={{ textDecoration: "none" }}
             >
-              <Button size="medium" startIcon={<Add />} variant="contained">
+              <Button
+                size="medium"
+                variant="contained"
+                startIcon={<Add />}
+                onClick={() => {
+                  logToMixpanel("IO_BO_SERVICE_CREATE_START", {});
```
giuseppedipinto, line 477: is it possible to refine the `logToMixpanel` function signature so that it doesn't have to pass empty object if there is nothing to specify?

apps/backoffice/src/providers/app.tsx
```diff
@@ -17,20 +19,31 @@ interface AppProviderProps {
...
-      </DrawerProvider>
-    </DialogProvider>
-  </ThemeProvider>
-);
+export const AppProvider = ({ children }: AppProviderProps) => {
+  //Mixpanel Initialization
+  mixpanel.init(getConfiguration().BACK_OFFICE_MIXPANEL_TOKEN, {
+    debug: true,
+    verbose: true,
+    track_pageview: false,
```
giuseppedipinto, line 27: Create a function on mix-panel.ts dedicated utility file and then call it here.

apps/backoffice/src/components/api-keys/api-key-value.tsx
```diff
@@ -7,12 +7,15 @@ import { LoaderSkeleton } from "../loaders";
 export interface ApiKeyValueProps {
   isVisible: boolean;
   keyValue?: string;
+  handleMixpanel?: () => void;
```
giuseppedipinto, line 10: please check and correct all code_review lint failures

giuseppedipinto (dismissed):

apps/backoffice/src/pages/services/[serviceId]/edit-service.tsx
Comment (carlo-bussi) [reply apps/backoffice/src/pages/services/[serviceId]/edit-service.tsx]: [fixed](https://github.com/pagopa/io-services-cms/pull/1018/commits/f19b230da57c7cb7ec1e72c997f7c35dec8f4ae9)

apps/backoffice/src/providers/app.tsx
carlo-bussi: [Done](https://github.com/pagopa/io-services-cms/pull/1018/commits/abc169ab602543d7a6658e84e1eba7c5ddafab6a)

giuseppedipinto: ok, clear old commented code 😉

apps/backoffice/src/pages/services/index.tsx
carlo-bussi: I tried to change the control to make possible to pass nothing if the event doesn't need any prop. but doing this it makes all events to have two kind of behaviour, anyone can accept both nothing or the proper pros that they needs

giuseppedipinto (approved):

PR #1030, opened 2024-10-01
Fix: modified_at overwrite on legacy_sync

giamma1295:
Add 1 milliseconds for publication record on Legacy -> CMS sync to avoid history overwrite
#### Checklist:


rodoviti (approved):
LGTM

PR #1037, opened 2024-10-08
[#IOPAE-1445] Internal GetServiceTopics

StefanoDoc:
New GetServiceTopicsInternal Api needed for Data ingestion 
#### List of Changes

1. New `GetServiceTopicsInternal` Api
2. New `InternalServiceTopics` type
3. New generic `findTopics` method in dao

- unit tests
- local start of the app

#### Checklist:


packages/io-services-cms-models/openapi/services-cms-schemas.yaml
```diff
@@ -321,3 +321,29 @@ ServiceTopic:
...
+    - id
+    - name
+    - deleted
+InternalServiceTopicList:
+  type: object
+  properties:
+    topics:
+      type: array
+      items:
+        $ref: '#/InternalServiceTopic'
```
rodoviti, line 349: **chore:** move these definitions directly to cms-webapp openapi.yaml

Because this is a "temporary" workaround to expose data to a specific internal stakeholder (my2cents: in this Use Case good design is to push data when it changes, so change the logic from pull to push), there is no need to put these "temporary schemas" on commons packages.

StefanoDoc: [5820204](https://github.com/pagopa/io-services-cms/pull/1037/commits/5820204e612de21762957ce817c8accfa2aa2f0f)

giamma1295 (approved):

PR #1044, opened 2024-10-14
[IOPAE-1351] get service handler return group name

rodoviti:
Resolve group name from its id and return it to the client

#### List of Changes

[retrive service list: resolve group name](https://github.com/pagopa/io-services-cms/commit/8a5314161d17aeb09bdfc14326a1c5d2bfef9863)

Unit Tests and local Dev Run

#### Checklist:


giamma1295 (approved):
LGTM

PR #1048, opened 2024-10-15
[#IOPAE-1457] Ingestion service publication

giamma1295:
Ingest the `services-publication` entity into the PDND data lake, using Eventhub.

#### List of Changes
- Add Generic Handlers to Ingest entities and for ingest retry
- Add Avro schema generator and validator `yarn generate` for `io-services-cms-webapp` will now generate classes, interfaces and so on from the avro schema
- Instantiate both ingest CosmosDBTrigger and the retry QueueTrigger for the entity `services-publication`
- created the DLQ `request-services-publication-ingestion-retry` in fn storage account

unit tests

#### Checklist:


apps/app-backend/api/internal.yaml
```diff
@@ -435,6 +435,9 @@ components:
               required:
                 - id
                 - name
+            group_id:
+              type: string
+              minLength: 1
```
rodoviti, line 440: **chore**: we actually don't need to propagate the incoming new `group_id` property to "App IO services storage".

`group_id` property will be removed before synching `app-backend` (db) from `db-services-cms`
```suggestion
            group_id:
              type: string
              minLength: 1
```

giamma1295: removed unintended changes in 51cbbcea7d9b624488367d4790a68d778192874e

rodoviti (approved):

Krusty93 (approved):
Looks good to me, but next time use managed identities instead of connection strings if possible :)

giamma1295:
> Looks good to me, but next time use managed identities instead of connection strings if possible :)

yup, we already have a jira task for it!

PR #1163, opened 2025-02-17
[IOPAE-1304] add postgres flexible servers for migration

rodoviti:
Add two postgres flexible server for migration from single server: one as a target and one as _Runtime Server_

#### Checklist:

### DIFF
--- a/infra/src/database_subsmigration.tf
+++ b/infra/src/database_subsmigration.tf
@@ -12,8 +12,8 @@ module "subsmigration_postgres_flexible_server" {
   delegated_subnet_id      = module.postgres_flexible_snet.id
 
   ### Admin
-  administrator_login    = data.azurerm_key_vault_secret.devportalservicedata_db_server_adm_username
-  administrator_password = data.azurerm_key_vault_secret.devportalservicedata_db_server_adm_password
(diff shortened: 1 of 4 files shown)


infra/src/database_subsmigration.tf
```diff
@@ -0,0 +1,54 @@
...
+
+  custom_metric_alerts = null
+  alerts_enabled       = true
+
+  diagnostic_settings_enabled = false
+
+  tags = var.tags
+}
+
+# resource "azurerm_postgresql_flexible_server_database" "subsmigration_database" {
```
mamu0, line 49: why is it commented?

rodoviti: Because the database should be created by the migration process (if I understand correctly, it should do a “dump-and-restore” behind the scenes). So, I would let the migration process create the DB (with the correct charset, collation, etc.), then add it to the `tf` state. Does that work for you?

leopoldo94 (approved):

infra/src/database_subsmigration.tf
mamu0: ah now it's clear, yes it's fine for me

PR #1169, opened 2025-02-20
[IOPAE-1849] backoffice fix service detail api

leopoldo94:
#### List of Changes
[fix service detail api](https://github.com/pagopa/io-services-cms/commit/e641147754f2382a10508c7e56f41969023c77d5)

#### Motivation and Context
a bug found when the user tries to call the service detail api and  is not an admin and belongs to at least one selfcare group

Unit test
Local environment connected to prod resources

#### Checklist:


rodoviti (approved):
lgtm

PR #1260, opened 2025-05-05
[IOPAE-2004] workaround override io organization name

leopoldo94:
#### List of Changes
[fix organization name override with temporary workaround](https://github.com/pagopa/io-services-cms/commit/23aa15bc9871aaecfa597973d7e5de1f637237ec)

#### Motivation and Context
When a service from the pagopa organization is updated or created the organazation name needs to be remapped 

Local environment connected to prod resources

#### Checklist:

### DIFF
--- a/apps/io-services-cms-webapp/src/utils/converters/service-lifecycle-converters.ts
+++ b/apps/io-services-cms-webapp/src/utils/converters/service-lifecycle-converters.ts
@@ -3,6 +3,7 @@ import {
   IResponseErrorInternal,
   ResponseErrorInternal,
 } from "@pagopa/ts-commons/lib/responses";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import * as TE from "fp-ts/lib/TaskEither";
 import { pipe } from "fp-ts/lib/function";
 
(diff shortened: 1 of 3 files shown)


apps/io-services-cms-webapp/src/utils/converters/service-lifecycle-converters.ts
```diff
@@ -111,6 +119,19 @@ const toServiceTopic =
         )
       : TE.right(undefined);
 
+//FIXME: remove this workaround for managing the creation and the update of services from pagopa
+//        where the name of the organization needs to be renamed into "IO - L'app dei servizi pubblici"
+const IO_APP_ORGANIZATION_NAME = "IO - L'app dei servizi pubblici";
```
giuseppedipinto, line 124: Consider to rename this var into `PAGOPA_ORGANIZATION_NAME`

apps/io-services-cms-webapp/src/utils/converters/service-lifecycle-converters.ts
```diff
@@ -111,6 +119,19 @@ const toServiceTopic =
...
+const PAGOPA_ORGANIZATION_FISCAL_CODE = "15376371009";
+
+const renameIoAppServicesOrganizationName = (
```
giuseppedipinto, line 127: Consider to rename this fn into `renamePagoPAServicesOrganizationName` because, imo, the concept is "I rename the org.name for PagoPA services"

giuseppedipinto (changes requested):
Additionally consider having a couple of unit tests for specific cases.

rodoviti (commented):
Since the problem is related to BackOffice, I would prefer to make the change there, specifically in the two places where this value is passed to CMS, which are:
-  _CreateService_ (`apps/backoffice/src/app/api/services/route.ts`)
- _EditService_ (`apps/backoffice/src/app/api/services/[serviceId]/route.ts`)

PR #1261, opened 2025-05-06
[IOPAE-2004] backoffice workaround override io organization name

leopoldo94:
#### List of Changes
[fix organization name override with temporary workaround](https://github.com/pagopa/io-services-cms/commit/4579775557a4478c271dbd25678adf8f7b1585b2)

When a service from the pagopa organization is updated or created the organazation name needs to be remapped

Local environment connected to prod resources and unit tests

#### Checklist:

### DIFF
--- a/apps/backoffice/src/utils/rename-pagopa-organization.ts
+++ b/apps/backoffice/src/utils/rename-pagopa-organization.ts
@@ -1,0 +1,12 @@
+//FIXME: remove this workaround for managing the creation and the update of services from pagopa
+//        where the name of the organization needs to be renamed into "IO - L'app dei servizi pubblici"
+const PAGOPA_ORGANIZATION_NAME = "IO - L'app dei servizi pubblici";
+const PAGOPA_ORGANIZATION_FISCAL_CODE = "15376371009";
+
+export const renamePagoPAServicesOrganizationName = (
+  fiscalCode: string,
+  organizationName: string,
+): string =>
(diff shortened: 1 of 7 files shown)


giuseppedipinto (approved):

apps/backoffice/src/utils/rename-pagopa-organization.ts
```diff
@@ -0,0 +1,12 @@
+//FIXME: remove this workaround for managing the creation and the update of services from pagopa
+//        where the name of the organization needs to be renamed into "IO - L'app dei servizi pubblici"
+const PAGOPA_ORGANIZATION_NAME = "IO - L'app dei servizi pubblici";
+const PAGOPA_ORGANIZATION_FISCAL_CODE = "15376371009";
+
+export const renamePagoPAServicesOrganizationName = (
```
rodoviti, line 6: **suggestion**: use a more generic name for the function.

The output of this function is to return the organization name, in addition it apply a name mapping in a specific case (when the org is PagoPA). Using `renamePagoPAServicesOrganizationName` as fn name could be could be misleading, 
I suggest modifying the function so that it takes as input a single parameter of type BackofficeUser (or better BackofficeUser["institution"] type ) and return the organization name (such as`getServiceOrganizationName(backofficeUser BackofficeUser): string`).
This way, when and if we remove this workaround, we only have to change the logic of the `getServiceOrganizationName` fn, with no impact on callers

rodoviti (approved):

PR #1265, opened 2025-05-16
add devcontainer config

rodoviti:
Add basic devcontainer config

#### List of Changes

- ddd basic devcontainer config

#### Checklist:

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,7 +2,6 @@
   "name": "io-services-cms",
   "image": "mcr.microsoft.com/devcontainers/base:debian@sha256:6155a486f236fd5127b76af33086029d64f64cf49dd504accb6e5f949098eb7e",
   "features": {
-    "ghcr.io/devcontainers-contrib/features/pre-commit:2": {},
     "ghcr.io/devcontainers/features/azure-cli:1": {},
     "ghcr.io/devcontainers/features/github-cli:1": {},
     "ghcr.io/devcontainers/features/terraform:1": {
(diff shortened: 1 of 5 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,55 @@
...
+        "vitest.explorer",
+        "redhat.vscode-yaml",
+        "dbaeumer.vscode-eslint",
+        "github.copilot",
+        "github.copilot-chat"
+      ],
+      "settings": {
+        "editor.formatOnSave": true,
+        "editor.defaultFormatter": "esbenp.prettier-vscode"
+      }
```
lucacavallaro, line 52: I suggest to move these settings to `.vscode/settings.json`, to make them also available for those who don't use devcontainers

Don't forget to configure also the `defaultFormatter` for terraform files (https://github.com/pagopa/io-messages/blob/main/.vscode/settings.json#L5), the extension needed is included automatically by the `ghcr.io/devcontainers/features/terraform:1` feature

If you need `func`, add our feature https://github.com/pagopa/dx/tree/main/.devcontainer/features/azure-functions-core-tools instead of the official one, that is not compatible with apple silicon

leopoldo94:
@rodoviti The dev dependency `"azure-functions-core-tools": "^4.0.7030"` inside the service-cms-webapp should be removed. When I try to run the function using Yarn, this dependency seems to take priority over the package installed in the container because it looks for executables in `node_modules` first. This is causing a Rosetta translation `error: rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2`.

.devcontainer/devcontainer.json
rodoviti: With the addition of `azure-functions-core-tools` feature, node scripts using `func` fail due to the presence of `“azure-functions-core-tools”:“^4.0.7030”` as the project's devDependency, which takes precedence over the globally installed one.
We introduced it as devDependency to centralize the dependency with a self-contained cross-platform package....do you know a way to "force" the container to use the system globally installation instead of the one with devDependency? Otherwise I think we should remove that devDependency in order to use DevContainer, correct?

lucacavallaro: Just remove that `dev dependency`, the best way to install `func` is by `brew` on macOS and system PM on linux.

Even though it might seem convenient to include it among those dependencies, it not only creates "priority" issues within the system, but it also significantly slows down our CI/CD workflows, as it's a very heavy dependency.

leopoldo94 (approved):

PR #1292, opened 2025-06-16
patch: try to fix cms opex gha

rodoviti:
#### Checklist:

### DIFF
--- a/.github/workflows/opex_api_services_cms.yml
+++ b/.github/workflows/opex_api_services_cms.yml
@@ -13,7 +13,6 @@ on:
 
 env:
   API_NAME: api_services_cms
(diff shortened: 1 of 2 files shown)


leopoldo94 (approved):

.github/workflows/opex_api_services_cms.yml
```diff
@@ -13,6 +13,10 @@ on:
 
 env:
   API_NAME: api_services_cms
+  ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
```
Krusty93, line 16: The `ARM_CLIENT_ID` is set at environment scope and not at the entire repository, so it should be moved to line 29 to make it works

Krusty93 (approved):

PR #1306, opened 2025-07-11
[IOPAE-2098] update ingestion function

leopoldo94:
#### List of Changes
[added new model for activation](https://github.com/pagopa/io-services-cms/commit/3f4120fc051429504a5631c22d3dda3554130c36)
[new blob trigger function](https://github.com/pagopa/io-services-cms/commit/a37e8a10a2f1aa573c6a672fd2b78ee3e6e8a6dc)
[implementation of the handler for the blob trigger](https://github.com/pagopa/io-services-cms/commit/b3a79ff1018008dfddada20170bf792a32ac4a4f)
[Remove obsolete retry function](https://github.com/pagopa/io-services-cms/commit/8cbd7244bcebacc1090c41f26c4c5c8c9242d25c)
[added unit tests](https://github.com/pagopa/io-services-cms/commit/9ede18c8d5c531d8b74ede0d2f281833865d6add)
[infra ingestion activation](https://github.com/pagopa/io-services-cms/commit/c52ab2681f13c3af18086cab9a0c3e8bc8662bf3)

#### Motivation and Context
We have replaced Cosmos DB with Blob Storage for storing data, which also involved updating the data model.

This change means we no longer need the Cosmos DB change feed. Instead, a Blob Trigger function now ingests the data. This trigger has its own built-in logic for retrying failed processing and sending messages to a poison queue, so our custom error-handling queues were removed.

Local environment and unit test

#### Checklist:


apps/io-services-cms-webapp/src/watchers/on-activation-ingestion-change.ts
```diff
@@ -16,13 +21,64 @@ export const handler = (
...
+export const parseBlob =
+  () =>
+  <R>(
+    processItems: RTE.ReaderTaskEither<
+      { items: Activations.Activation[] },
+      Error,
+      R
+    >,
+  ): RTE.ReaderTaskEither<AzureFunctionCall, Error, R> =>
+  ({ context, inputs }) => {
```
rodoviti, line 44: **nitpick:** add return type value to function definition

apps/io-services-cms-webapp/src/utils/ingestion/__tests__/ingestion-handlers.test.ts
```diff
@@ -130,4 +132,48 @@ describe("Generic Ingestion PDND Handlers", () => {
...
+
+    it("should throw on eventhub error", async () => {
+      //given when
+      const res = await createIngestionBlobTriggerHandler(
+        aProducerWhichFails,
+        aFormatter,
+      )({ items: [anItem] })();
+
+      //then
+      expect(E.isLeft(res)).toBeTruthy();
```
rodoviti, line 161: **chore:** you should also verify that the actual value of E.left param is equals to the expected one (for each test case in which the handler return a left object)

apps/io-services-cms-webapp/src/utils/ingestion/__tests__/ingestion-handlers.test.ts
```diff
@@ -130,4 +134,57 @@ describe("Generic Ingestion PDND Handlers", () => {
...
+
+    it("should throw on eventhub error", async () => {
+      //given when
+      const res = await createIngestionBlobTriggerHandler(
+        aProducerWhichFails,
+        aFormatter,
+      )({ items: [anItem] })();
+
+      //then
+      if (E.isLeft(res)) {
```
rodoviti, line 165: **chore:** you have to add `expect(E.isLeft(res)).toBeTruthy();` before narrowing the Either value to verify its value to avoid "false positive" results when the test case return in an Either.right value.....in this case the test will pass without error, but it is not what we expect.
```suggestion
      expect(E.isLeft(res)).toBeTruthy();
      if (E.isLeft(res)) {
```

apps/io-services-cms-webapp/src/watchers/on-activation-ingestion-change.ts
```diff
@@ -16,13 +21,60 @@ export const handler = (
...
+  >,
+) => RTE.ReaderTaskEither<AzureFunctionCall, Error, R> =
+  () =>
+  <R>(
+    processItems: RTE.ReaderTaskEither<
+      { items: Activations.Activation[] },
+      Error,
+      R
+    >,
+  ): RTE.ReaderTaskEither<AzureFunctionCall, Error, R> =>
```
rodoviti, line 51: **suggestion:** since this function has no input dependencies, you could “simplify” it as suggested
```suggestion
export const parseBlob: () => <R>(
  processItems: RTE.ReaderTaskEither<
    {
      items: Activations.Activation[];
    },
    Error,
    R
  >,
) => RTE.ReaderTaskEither<AzureFunctionCall, Error, R> =
  () =>
  <R>(
    processItems: RTE.ReaderTaskEither<
      { items: Activations.Activation[] },
      Error,
      R
    >,
  ): RTE.ReaderTaskEither<AzureFunctionCall, Error, R> =>
```

leopoldo94: fixed with c436b2a69b9f34642f5375d636046801cc00845b

leopoldo94: fixed with dfb1daac462723e2a0d0190372ec6881c5854e80

apps/io-services-cms-webapp/src/watchers/on-activation-ingestion-change.ts
```diff
@@ -16,13 +21,68 @@ export const handler = (
...
+                  `Failed parsing the blob [${readableReport(errors)}]`,
+                ),
+            ),
+          ),
+        ),
+      ),
+      TE.fromEither,
+      TE.chainW((decodedItem) => processItems({ items: [decodedItem] })),
+    );
+  };
```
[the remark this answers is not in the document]
leopoldo94: fixed with c436b2a69b9f34642f5375d636046801cc00845b

rodoviti (approved):

PR #1313, opened 2025-07-17
fix(form create/update service): added required fields where necessary

dcaccamo3ainformatica:
#### List of Changes

This PR introduces a minor improvement focused on the proper handling of the required property in the form components used within the service creation/update process.
Specifically:
Explicit required flags have been added to **SelectController**, **TextFieldController**, and **AutocompleteController** where required by validation.
In the **ServiceGroupSelector**, the group selection field now receives the required prop dynamically based on user permissions and active configurations.
No business logic or flow changes have been made—this is strictly a UI/validation alignment fix.

#### Motivation and Context

1. Better UI clarity for required fields
2. Consistency between UI validation and schema validation logic
3. Proper handling of group_id field requirements for operators with enabled groups

Manual verification of the forms in:

1. Service creation and update flows
2. Group selection based on session context
3. Validation checks with both operator and admin roles
4. MSW mock and local flows verified

P.S I added the appropriate token for an operator with enabled groups, this in the mock is essential to test the functionality with the right operator. Obviously I think the old token should be restored once the functionality has been tested, when in doubt I have also committed this change to the mock

https://github.com/user-attachments/assets/7f189b78-0e8f-449f-a4a1-d9fe3f1dee0e

#### Checklist:

- [X] My change requires a change to the documentation.


apps/backoffice/src/components/services/service-create-update/service-builder-step-3.tsx
```diff
@@ -67,6 +69,11 @@ export const ServiceBuilderStep3 = ({
   const isGroupSelectorVisible = () =>
     mode === "create" && groups && groups?.length > 0;
 
+  const isGroupRequired =
```
giuseppedipinto, line 72: Why not create a utility function exported from the `src/utils/auth-util.ts` file and reuse it wherever needed? _(I believe this check is already done in one or more places in the code)_.

giuseppedipinto (changes requested):

apps/backoffice/src/components/services/service-create-update/service-builder-step-3.tsx
dcaccamo3ainformatica: bcbd5f1c

giuseppedipinto (approved):

PR #1315, opened 2025-07-21
[IOPAE-1895] ApiKeyGroup Bulk association: improved user interaction on modal

dcaccamo3ainformatica:
#### List of Changes

Added “Select All” and “Deselect All” on ApikeyGroup Bulk association on modal, and checkbox on all item service on modal

#### Motivation and Context
added checkboxes and logic for mass selection of services to be associated to improve the user experience

Local dev mode with msw mocks.

#### Screenshots

 (if appropriate):

#### Checklist:

### DIFF
--- a/apps/backoffice/src/components/forms/controllers/multi-select-controller.tsx
+++ b/apps/backoffice/src/components/forms/controllers/multi-select-controller.tsx
@@ -1,7 +1,6 @@
 import { LoaderSkeleton } from "@/components/loaders";
 import {
   Box,
-  Button,
   Checkbox,
   Chip,
   Divider,
@@ -14,8 +13,9 @@ import {
   Select,
   SelectProps,
 } from "@mui/material";
+import { ButtonNaked } from "@pagopa/mui-italia";
 import { useTranslation } from "next-i18next";
-import React, { ReactNode } from "react";
+import { ReactNode } from "react";
 import { Controller, get, useFormContext } from "react-hook-form";
 
 export type MultiSelectControllerProps = {
@@ -37,15 +37,14 @@ const getMultiSelectLogic = (
   const allSelected = selected.length === items.length;
 
   const toggleItem = (itemValue: number | string) => {
-    const stringVal = String(itemValue);
-    const newSelected = selected.includes(stringVal)
-      ? selected.filter((v) => v !== stringVal)
-      : [...selected, stringVal];
+    const newSelected = selected.includes(itemValue)
+      ? selected.filter((v) => v !== itemValue)
+      : [...selected, itemValue];
     onChange(newSelected);
   };
 
   const handleSelectAll = () => {
-    const allValues = items.map((i) => String(i.value));
+    const allValues = items.map((i) => i.value);
     onChange(allSelected ? [] : allValues);
   };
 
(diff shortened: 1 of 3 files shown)


apps/backoffice/src/components/forms/controllers/multi-select-controller.tsx
```diff
@@ -27,76 +32,117 @@ export type MultiSelectControllerProps = {
...
-                  flexWrap: "wrap",
-                  gap: 0.5,
-                }}
-              >
-                {selected.map((value: React.Key | null | undefined) => (
-                  <Chip
-                    key={value}
-                    label={items.find((item) => item.value === value)?.label}
+      render={({ field: { onChange, value = [] } }) => {
+        const selected = Array.isArray(value) ? value : [];
```
giuseppedipinto, line 56: All the logic _(toggleItem, handleSelectAll, handleDeselectAll, calculations for selected, allSelected, etc.)_ is written inline in the Controller render, resulting in the following critical issues:

- Difficult to read: everything is flattened in the return, making the component longer and less readable.
- Difficult to test: the functions are anonymous and cannot be tested in isolation.
- Difficult to debug: especially during development or refactoring.
- Difficult to re-render: each re-render of the component creates these functions again, which can have a performance impact _(although minor)._

Try to avoid this by moving the logic outside the return and using the return only for the declarative part of the JSX.

apps/backoffice/src/components/groups/associate-groups-create-update/associate-groups-builder-step.tsx
```diff
@@ -70,6 +72,15 @@ export const AssociateGroupsBuilderStep = ({
               : []
           }
           label={t("forms.groups.associate.services.select.placeholder")}
+          labels={{
```
giuseppedipinto, line 75: I don't think it's useful to expose this parameter _(it's also a bit misleading with the existing `SelectProps` "label")_.
Configure the labels directly within the **MultiSelectController** component without having to pass them in from outside.

dcaccamo3ainformatica: OK I will make the change

apps/backoffice/src/components/forms/controllers/multi-select-controller.tsx
dcaccamo3ainformatica: In fact all the logic is left in the Controller render and this compromises the clarity and extensibility of the component.
I proceed with a refactor by moving the selection management functions (toggleItem, selectAll, etc.) out of the render.

apps/backoffice/src/components/forms/controllers/multi-select-controller.tsx
```diff
@@ -27,76 +32,117 @@ export type MultiSelectControllerProps = {
...
-            {error ? error.message : (helperText ?? null)}
-          </FormHelperText>
-        </FormControl>
-      )}
+                </MenuItem>
+              )}
+
+              {showDeselect && (
+                <MenuItem key="deselect-all" onClick={handleDeselectAll}>
+                  <Checkbox checked={false} />
```
giuseppedipinto, line 120: The **_Select/Deselect All_** layout is not the one shown in _Figma_, which requires a single action button at a time rendered as a Text Button (without checkbox)

apps/backoffice/src/components/forms/controllers/multi-select-controller.tsx
```diff
@@ -27,76 +32,117 @@ export type MultiSelectControllerProps = {
...
+            >
+              {items.length > 1 && (
+                <MenuItem key="select-all" onClick={handleSelectAll}>
+                  <Checkbox
+                    checked={allSelected}
+                    indeterminate={isIndeterminate}
+                  />
+                  <ListItemText
+                    primary={labels.selectAll}
+                    sx={{ color: "#0073E6" }}
```
giuseppedipinto, line 113: Please avoid direct ref to hex color _(what would happen if it changed in the mui-italia library?)_. In this case you should be able to use `color="primary"` prop for **ListItemText**.

apps/backoffice/src/components/groups/associate-groups-create-update/associate-groups-builder-step.tsx
dcaccamo3ainformatica: 584ce4c5

apps/backoffice/src/components/forms/controllers/multi-select-controller.tsx
dcaccamo3ainformatica: 1903dbd9

PR #1325, opened 2025-07-25
[#IOPAE-2121] BFF - Definition of openapi template of get user institution products api

Iustin-Burlacu:
#### List of Changes

added in openapi template of the backoffice the definition of get user institution products api

#### Motivation and Context

required new api from backoffice portal to show the list of institutional active products

in local

#### Checklist:

### DIFF
--- a/apps/backoffice/openapi.yaml
+++ b/apps/backoffice/openapi.yaml
@@ -1024,6 +1024,37 @@ paths:
           description: Too many requests
         '500':
           description: Internal server error
+  /institutions/{institutionId}/products:
+    get:
+      tags:
+        - user-products-institutions
+      summary: Retrieve all active institutional products of the user
+      description: Retrieve all active institutional products for the given user and institution
+      operationId: getUserInstitutionProducts
+      security:
+        - cookieAuth: []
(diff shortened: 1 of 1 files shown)


apps/backoffice/openapi.yaml
```diff
@@ -1024,6 +1024,37 @@ paths:
           description: Too many requests
         '500':
           description: Internal server error
+  /institutions/{institutionId}/products:
+    get:
+      tags:
+        - user-products-institutions
```
rodoviti, line 1030: **chore**: use `institutions` instead of `user-products-institutions` tag.

We use tag to logically group APIs, so in this case we have groped this type of APIs based on their "main" entity: Institution
```suggestion
        - institutions
```

rodoviti (changes requested):

apps/backoffice/openapi.yaml
Iustin-Burlacu: @rodoviti Fixed with f2bc1a37a76b439a1d70307ecc3d18e909dbe780

giuseppedipinto (approved):

PR #1328, opened 2025-07-29
[#IOPAE-2122] - FE Change logic load more on api key of group

dcaccamo3ainformatica:
#### List of Changes

change logic load more on api key of group

#### Motivation and Context

we don't have an total element from be

in local with mock with msw and simulate 21 element or 20 

#### Checklist:

### DIFF
--- a/apps/backoffice/src/components/api-keys/api-keys-groups/api-keys-groups.tsx
+++ b/apps/backoffice/src/components/api-keys/api-keys-groups/api-keys-groups.tsx
@@ -341,8 +341,8 @@ export const ApiKeysGroups = (props: ApiKeysGroupsProps) => {
         >
           <></>
         </LoaderSkeleton>
-        {mspData &&
-          mspData?.pagination.count > pagination.offset + pagination.limit && (
+        {Array.isArray(mspData?.value) &&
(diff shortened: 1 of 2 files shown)


apps/backoffice/src/components/api-keys/api-keys-groups/api-keys-groups.tsx
```diff
@@ -341,8 +341,8 @@ export const ApiKeysGroups = (props: ApiKeysGroupsProps) => {
         >
           <></>
         </LoaderSkeleton>
-        {mspData &&
-          mspData?.pagination.count > pagination.offset + pagination.limit && (
+        {Array.isArray(mspData?.value) &&
```
giuseppedipinto, line 344: Is the `isArray` check necessary?
TypeScript's typecheck already correctly models the value as an array of objects, and `useFetch` decodes the data using the codec specified in the fetch usage definition, in this case, `SubscriptionPagination`.
Here `mspData` can be _undefined_ or an object in the SubscriptionPagination format.
Therefore, I think it's enough to simplify with:
`{ mspData?.value.length === pagination.limit && (`
nothing else.

giuseppedipinto (changes requested):

PR #1331, opened 2025-07-31
[IOPAE-2003] Removed category and custom_special_flow from openapi of cms

Iustin-Burlacu:
#### List of Changes

removed category and custom_special_flow from openapi of cms and fix the related tests

#### Motivation and Context

following the fix for a bug in which the category and custom_special_flow fields of special services were overwritten, for a cleaner code, the aforementioned fields should be removed from the CMS OpenAPI.

unit test

#### Checklist:


rodoviti (approved):
lgtm

PR #1332, opened 2025-07-31
[IOPAE 2003] Patch removed category and custom special flow from openapi of cms patch

Iustin-Burlacu:
#### List of Changes

removed category and custom_special_flow from cms and backoffice and fix the related tests

#### Motivation and Context

following the fix for a bug in which the category and custom_special_flow fields of special services were overwritten, for a cleaner code, the aforementioned fields should be removed from the CMS OpenAPI.

unit test

#### Checklist:


apps/io-services-cms-webapp/src/utils/converters/service-lifecycle-converters.ts
```diff
@@ -41,7 +38,7 @@ export const payloadToItem = (
       sandboxFiscalCode,
     ),
     max_allowed_payment_amount,
-    metadata: { ...metadata, category: toCategoryType(metadata.category) },
+    metadata: metadata,
```
rodoviti, line 41: **nitpick**: use short-form
```suggestion
    metadata,
```

apps/io-services-cms-webapp/src/webservice/controllers/create-service.ts
```diff
@@ -119,8 +118,6 @@ export const makeCreateServiceHandler =
                   ...serviceItem.data,
                   metadata: {
                     ...serviceItem.data.metadata,
-                    category: CategoryEnum.STANDARD,
```
leopoldo94, line 122: Issue: if we remove this piece, we will allow the creation of services with the category `undefined`. 
Suggestion: Maybe a solution could be to create a function like this 

`const getStandardCategory =
  (): ServiceLifecycle.definitions.Service["data"]["metadata"]["category"] =>
    "STANDARD";`

and substitute this 

`category: getStandardCategory(),`

 to not rely on the definition of `CategoryEnum` created by the openapi

apps/io-services-cms-webapp/src/utils/converters/service-publication-converters.ts
```diff
@@ -21,7 +20,7 @@ export const itemToResponse =
   ({
     data: {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      metadata: { category, custom_special_flow, scope, topic_id, ...metadata },
```
leopoldo94, line 24: Issue: This destructuring needs to remain as is because it prevents the fields `category` and `custom_special_flow` from being returned in the response.

Iustin-Burlacu: @leopoldo94 fixed with 28286059743a0c46187e29ff6bbc87befaa9055f

apps/io-services-cms-webapp/src/utils/converters/service-lifecycle-converters.ts
```diff
@@ -60,7 +57,7 @@ export const itemToResponse =
   ({
     data: {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      metadata: { category, custom_special_flow, scope, topic_id, ...metadata },
```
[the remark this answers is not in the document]
Iustin-Burlacu: @leopoldo94 fixed with 28286059743a0c46187e29ff6bbc87befaa9055f

apps/io-services-cms-webapp/src/webservice/controllers/create-service.ts
```diff
@@ -83,6 +82,10 @@ const createSubscription = (
     ),
   );
 
+const getStandardCategory =
+  (): ServiceLifecycle.definitions.Service["data"]["metadata"]["category"] =>
+    "STANDARD";
```
rodoviti, line 87: **nitpick**: use a constant instead of a function
```suggestion
const STANDARD_CATEGORY: ServiceLifecycle.definitions.Service["data"]["metadata"]["category"] =
  "STANDARD";
```

Iustin-Burlacu: @rodoviti fixed with 998c8914307e618725ff38e041813b8ff3ae3526

PR #1338, opened 2025-08-07
patch roles for activations blob storage

leopoldo94:
Change the blob storage role for the cms function app after the devex module update in which the writer role now has the permission to write tags on blobs

#### Checklist:


rodoviti (approved):
lgtm

PR #1347, opened 2025-09-01
Fix switch to Reserved Area forcing BackOffice IO logout

giuseppedipinto:
Fix the switch to Reserved Area, forcing BackOffice IO logout.

This PR addresses a bug that occurred when a user switched from the BackOffice IO to the Reserved Area, changed their institution, and then returned to the BackOffice.
Previously, this sequence of actions could lead to the incorrect institution being displayed.
This fix resolves the issue by forcing a BackOffice IO logout whenever a user switches to the Reserved Area, ensuring that the correct entity is always shown upon their return.

Closes IOPAE-2155 Resolves IOPAE-2156

### DIFF
--- a/apps/backoffice/src/components/headers/header.tsx
+++ b/apps/backoffice/src/components/headers/header.tsx
@@ -62,7 +62,9 @@ export const Header = () => {
     // navigate to product url (i.e.: SelfCare)
     else if (product.id === SELFCARE_ID) {
       setSelectedProductId(product.id);
-      window.location.href = product.productUrl;
+      signOut({
+        callbackUrl: product.productUrl,
+      });
     }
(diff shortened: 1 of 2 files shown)


apps/backoffice/src/components/headers/header.tsx
```diff
@@ -62,7 +62,9 @@ export const Header = () => {
     // navigate to product url (i.e.: SelfCare)
     else if (product.id === SELFCARE_ID) {
       setSelectedProductId(product.id);
-      window.location.href = product.productUrl;
+      signOut({
+        callbackUrl: product.productUrl,
+      });
```
rodoviti, line 67: **suggestion (non-blocking):** We could use `handleTokenExchange` function to centralize the logic

PR #1355, opened 2025-09-09
Save only the last comment of a Jira ticket

Iustin-Burlacu:
Save only the last comment of a Jira ticket if present

Resolves IOPAE-2160


giuseppedipinto (approved):
LGTM

PR #1468, opened 2026-02-26
Pin hash commit for slack notification action in Selfcare OpenAPI drift detection

Iustin-Burlacu:
Pin hash commit for slack notification action in Selfcare OpenAPI drift detection
[On the fly]


rodoviti:
closed because, in accordance with our organization's policy, pinning an internal (DX) dependency it is not mandatory

PR #1540, opened 2026-05-12
feat: add `isAdminAggregator` role check and corresponding tests

salvatorelaiso:
Resolves [[#IOPID-3922]](https://pagopa.atlassian.net/browse/IOPID-3922), [[#IOPID-3923]](https://pagopa.atlassian.net/browse/IOPID-3923)

### DIFF
--- a/apps/backoffice/src/app/api/subscriptions/[subscriptionId]/keys/[keyType]/handler.ts
+++ b/apps/backoffice/src/app/api/subscriptions/[subscriptionId]/keys/[keyType]/handler.ts
@@ -12,6 +12,7 @@ import {
 import { sanitizedNextResponseJson } from "@/lib/be/sanitize";
 import { regenerateManageSubscriptionApiKey } from "@/lib/be/subscriptions/business";
 import { BackOfficeUserEnriched } from "@/lib/be/wrappers";
+import { ApimUtils } from "@io-services-cms/external-clients";
 import { readableReport } from "@pagopa/ts-commons/lib/reporters";
 import * as E from "fp-ts/lib/Either";
 import { NextRequest, NextResponse } from "next/server";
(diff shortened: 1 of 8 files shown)


apps/backoffice/src/app/api/subscriptions/[subscriptionId]/keys/[keyType]/handler.ts
```diff
@@ -26,7 +27,15 @@ export const regenerateManageSubscriptionKeyHandler = async (
...
   },
 ): Promise<NextResponse<ResponseError | SubscriptionKeys>> => {
-  if (!userAuthz(backofficeUser).isAdmin()) {
+  const allowed =
+    userAuthz(backofficeUser).isAdmin() ||
+    userAuthz(backofficeUser).isAnAggregatorAdminAllowedOnGroup(
+      params.subscriptionId.substring(
+        ApimUtils.SUBSCRIPTION_MANAGE_GROUP_PREFIX.length,
+      ),
+    );
```
Comment (rodoviti) [inline apps/backoffice/src/app/api/subscriptions/[subscriptionId]/keys/[keyType]/handler.ts:36]: **question:** am I wrong, or is the new function `isAnAggregatorAdminAllowedOnGroup` actually useless, since this check can be rewritten using `isGroupAllowed` fn?!
```suggestion
    userAuthz(backofficeUser).isAdmin() ||
    (userAuthz(backofficeUser).isAggregatorAdmin() &&
    userAuthz(backofficeUser).isGroupAllowed(
      params.subscriptionId.substring(
        ApimUtils.SUBSCRIPTION_MANAGE_GROUP_PREFIX.length,
      ),
    ));
```

Comment (salvatorelaiso) [reply apps/backoffice/src/app/api/subscriptions/[subscriptionId]/keys/[keyType]/handler.ts:36]: While this check can be rewritten that way, my goal here is to have a single function that checks a single case, unlike the current isGroupAllowed, which handles multiple cases (e.g. an admin or a user without any group).
Handling roles with the current approach is difficult and unclear for a first-time reader.

Even worse, reusing that function for this case would make its purpose even less clear and make it harder to follow the logic path that leads to permissions being granted in a specific scenario.

My aim here is to keep the current isGroupAllowed for now, while planning a later refactor of the checks by introducing a dedicated function for each role (admin, aggregator admin, operator).

Comment (rodoviti) [reply apps/backoffice/src/app/api/subscriptions/[subscriptionId]/keys/[keyType]/handler.ts:36]: makes sense. The purpose of the `isGroupAllowed` fn was to provide a single function to verifying proper authorization for a specific group, regardless of the user's role, but things have changed now! So I agree with you to keep the current isGroupAllowed for now and planning a refactor to improve this checks 😉

Comment (salvatorelaiso) [reply apps/backoffice/src/app/api/subscriptions/[subscriptionId]/keys/[keyType]/handler.ts:36]: At this stage, we could already introduce an `isUserAllowedOnGroup` function to handle the logic shared across these functions and avoid code duplication.
I’ll provide a commit for this soon.

Comment (salvatorelaiso) [reply apps/backoffice/src/app/api/subscriptions/[subscriptionId]/keys/[keyType]/handler.ts:36]: See https://github.com/pagopa/io-services-cms/pull/1540/commits/1c8a53c9474a7d1d54162353a7dbe47e372c0d7e

rodoviti (approved):

PR #1543, opened 2026-05-13
[#IOPID-3929] Custom prefix for manage group subscription display name

arcogabbo:


arcogabbo:
Closing due to different management system for special manage group subscription.

PR #1549, opened 2026-05-19
Update bootstrap apply workflow for prod

mamu0:
Adds the workflow to automatically apply the bootstrapper configuration.

### DIFF
--- a/.github/workflows/_release-terraform-apply-bootstrapper-prod.yaml
+++ b/.github/workflows/_release-terraform-apply-bootstrapper-prod.yaml
@@ -1,0 +1,20 @@
+name: Release Bootstrapper Infrastructure - prod
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - "infra/bootstrapper/prod/**"
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  release:
+    uses: pagopa/dx/.github/workflows/release-terraform-bootstrapper-v1.yaml@main
+    name: Release Bootstrapper (prod)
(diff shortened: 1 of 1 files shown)


rodoviti:
@mamu0 did you see [this](https://sonarcloud.io/project/security_hotspots?id=pagopa_io-services-cms&pullRequest=1549&hotspots=AZ4_VXxSDZAgQPchbbNC) SonarQube Security Hostspot alert? Is it safe to use?

mamu0:
> did you see [this](https://sonarcloud.io/project/security_hotspots?id=pagopa_io-services-cms&pullRequest=1549&hotspots=AZ4_VXxSDZAgQPchbbNC) SonarQube Security Hostspot alert? Is it safe to use?

@rodoviti Yes, the secrets it inherits are defined granularly within the workflow we set up on [dx](https://github.com/pagopa/dx/blob/43c272fb440af37f540fce8a44b043e1fd2fd932/.github/workflows/release-terraform-bootstrapper-v1.yaml#L28), so it doesn't have more permissions than it actually needs

rodoviti (approved):
lgtm

rodoviti (commented):
@mamu0 Why the GH workflow file starts with `_`? In this repo no other GH workflow files uses a naming convention with prefix...

mamu0:
> @mamu0 Why the GH workflow file starts with `_`? In this repo no other GH workflow files uses a naming convention with prefix...

It has this name because it follows the naming convention we use in dx for workflows
This keeps it consistent with the DX CLI, which automatically generates this workflow following the latest updates

PR #1551, opened 2026-05-21
[#IOPID-3909] Script to purge all aggregates starting from aggregator id

arcogabbo:


packages/utilities/src/scripts/purge-all-aggregates.ts
```diff
@@ -0,0 +1,221 @@
...
+          }
+          return TE.left(
+            new Error(`Failed deleting user ${userId}: ${JSON.stringify(err)}`),
+          );
+        }),
+      );
+    }),
+  );
+
+const PAGE_SIZE = 100;
```
rodoviti: **suggestion**: Maybe we could increase this parameter... In the worst-case scenario, we'd have to make about 80 API calls to retrieve all the delegations 😅

packages/utilities/src/scripts/purge-all-aggregates.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    if (E.isLeft(result)) {
+      logger.error(`ERROR purging ${institutionId}: ${result.left.message}`);
+      errorCount++;
+    } else {
+      logger.info(`Done: ${institutionId}`);
+      successCount++;
+    }
+  }
+
```
rodoviti, line 212: **chore**: I think we also have to delete/purge the aggregator

packages/utilities/src/scripts/purge-all-aggregates.ts
arcogabbo: done in https://github.com/pagopa/io-services-cms/pull/1551/commits/4d65295355589c7205d81228b1667aa0aebcfe5e

packages/utilities/src/scripts/purge-all-aggregates.ts
arcogabbo: I added an option (default is `false`) to also purge the aggregator in https://github.com/pagopa/io-services-cms/pull/1551/commits/7cf237e3a39bd46438f469160d455a7d54822cad . Please check the commit

rodoviti (approved):

PR #1558, opened 2026-05-22
[#IOPID-3942] Update release workflow 

salvatorelaiso:
Update release Action to use DX workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    paths:
+      - .changeset/**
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -3,103 +3,18 @@ name: Release
 on:
   push:
     branches:
-      - master
-    paths:
-      - .changeset/**
+      - main
```
rodoviti: **question**: why did you remove paths config? if I remember correctly, we added this filter because release can only occur if there is a change within that folder...this way we prevent the pipeline from running when it shouldn't
```suggestion
      - main
    paths:
      - .changeset/**
```

rodoviti (approved):

PR #1562, opened 2026-06-03
[#IOPID-3915] Add blob metrics tracking and alert for API keys export failures

salvatorelaiso:
Add metrics to track export status.
Add alert for failures.


arcogabbo (approved):
I guess infra related warning will be solved in another PR. Apart from those, the PR LGTM

rodoviti (approved):
lgtm

PR #1577, opened 2026-06-10
fix: filter expired blobs

salvatorelaiso:
Add a check to not retrieve expired blobs.

Resolves [IOPID-3995](https://pagopa.atlassian.net/browse/IOPID-3995)

[IOPID-3995]: https://pagopa.atlassian.net/browse/IOPID-3995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/backoffice/src/lib/be/__tests__/api-keys-exports-adapter.test.ts
+++ b/apps/backoffice/src/lib/be/__tests__/api-keys-exports-adapter.test.ts
@@ -284,139 +284,123 @@ describe("findExportsFiles", () => {
     },
   );
 
-  it("should skip blobs whose expiration date has already passed", async () => {
-    // given
-    const adapter = ApiKeysExportsAdapter.getInstance(environment);
-    // last modified more than EXPORTS_API_KEYS_DURATION_IN_HOURS (24h) before now
-    const hoursOffset = adapter.EXPORTS_API_KEYS_DURATION_IN_HOURS + 1;
-    const expiredDate = new Date(
-      mockDate.getTime() - hoursOffset * 60 * 60 * 1000,
-    );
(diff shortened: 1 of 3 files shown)


apps/backoffice/src/lib/be/__tests__/api-keys-exports-adapter.test.ts
```diff
@@ -277,6 +283,95 @@ describe("findExportsFiles", () => {
       expect(result).toStrictEqual(expectedResult);
     },
   );
+
+  it("should skip blobs whose expiration date has already passed", async () => {
```
arcogabbo: can we stress this check for all `FileStateEnum` values?

arcogabbo (approved):
LGTM, just a minor comment

apps/backoffice/src/lib/be/__tests__/api-keys-exports-adapter.test.ts
salvatorelaiso: Please, have a look at: https://github.com/pagopa/io-services-cms/pull/1577/commits/7dfae3e41e2db7164ed79762ca79d7836fcf2c4d

arcogabbo: thank you @salvatorelaiso

rodoviti (approved):
suggestion: (only for readability and maintainability improvement) I think you could write unit tests by using `it.each`, because the three tests you added do the same logic except for expected input and expected output, so they can be easily parameterized

PR #1579, opened 2026-06-10
[#IOPID-3909] manual retry capability for APIM listSecrets AbortError

arcogabbo:
Library treats AbortError as non retryable and doesn't trigger a retry with core-rest-pipeline integrated options. This PR swaps retry method with retries enforced (if options are present). This mechanism is only added to the listSecrets utility but can be extended to other use cases in the future

### DIFF
--- a/packages/external-clients/src/apim-client/index.ts
+++ b/packages/external-clients/src/apim-client/index.ts
@@ -34,9 +34,17 @@ import {
   SubscriptionKeyType,
   SubscriptionKeyTypeEnum,
 } from "../generated/api/SubscriptionKeyType";
+import { retryTaskEither } from "../utils/retries";
 import { subscriptionsExceptManageOneApimFilter } from "./apim-filters";
 
 export type ApimMappedErrors = IResponseErrorInternal | IResponseErrorNotFound;
+
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export type ApimClientCustomRetryOptions = {
+  readonly initialDelayMs?: number;
+  readonly maxDelayMs?: number;
+  readonly maxRetries?: number;
+};
 
 export const ApimRestError = t.intersection([
   t.type({
@@ -183,6 +191,7 @@ export const getApimService = (
   apimResourceGroup: string,
   apimServiceName: string,
   apimProductName: NonEmptyString,
+  retryOptions?: ApimClientCustomRetryOptions,
 ): ApimService => ({
   createGroupUser: (groupId, userId) =>
     createGroupUser(
@@ -272,6 +281,7 @@ export const getApimService = (
       apimServiceName,
       subscriptionId,
       requestTimeoutInMs,
+      retryOptions,
     ),
   regenerateSubscriptionKey: (subscriptionId, keyType) =>
     regenerateSubscriptionKey(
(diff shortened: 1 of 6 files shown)


packages/external-clients/src/utils/retries.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param maxRetries - Maximum number of retry attempts (0 = no retries).
+ * @param initialDelayMs - Initial delay in milliseconds before the first retry.
+ * @param maxDelayMs - Maximum delay cap in milliseconds.
+ * @param shouldRetry - Predicate to determine if the error is retriable.
+ */
+export const retryTaskEither = <E, A>(
+  action: TE.TaskEither<E, A>,
+  maxRetries: number,
+  initialDelayMs: number,
+  maxDelayMs: number,
```
rodoviti: **suggestion**: use a single custom type instead of n properties, like the following:

```
export type RetryOptions = {
  readonly initialDelayMs: number;
  readonly maxDelayMs: number;
  readonly maxRetries: number;
};
```

packages/external-clients/src/apim-client/index.ts
```diff
@@ -34,10 +34,27 @@ import {
...
+ * that are aborted by a client-side timeout: when a request times out, the SDK
+ * throws an AbortError immediately and skips all retry strategies. These options
+ * configure an application-level retry layer to fill that gap.
+ */
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export type ApimClientTimeoutRetryOptions = {
+  readonly initialDelayMs?: number;
+  readonly maxDelayMs?: number;
+  readonly maxRetries?: number;
+};
```
rodoviti: **suggestion**: make all three parameters required and so use the `RetryOptions` type defined in `retries.ts` or if you want a new one, create it as an alias of `RetryOptions`. In the second case, I suggest to not use ApimClient as prefix, because can be misleading. Instead, use `ApimServiceTimeoutRetryOptions` or simply `TimeoutRetryOptions`

PS: I think there is no need to export this type

packages/external-clients/src/apim-client/index.ts
```diff
@@ -449,21 +468,34 @@ const listSecrets = (
...
-    TE.tryCatch(
-      () =>
-        apimClient.subscription.listSecrets(
-          apimResourceGroup,
-          apimServiceName,
-          subscriptionId,
-          {
-            requestOptions: {
-              timeout: requestTimeoutInMs,
+    retryTaskEither(
```
rodoviti: **suggestion**: I think we can rewrite retryTaskEither in a "functional-composition" way in order to first define the TE action and then do something like that:
```
pipe(
   TE.tryCatch(...),
   retryOptions ? retryTaskEither(retryOptions, shouldRetry) : identity,
   ...
)
```

rodoviti (commented):
I left some comment to improve (I think) code readability.

packages/external-clients/src/utils/retries.ts
arcogabbo: refactor done in https://github.com/pagopa/io-services-cms/pull/1579/changes/4b6a53d9d20a05b23e648c8f3058eab98d6c6e29

packages/external-clients/src/apim-client/index.ts
arcogabbo: I went for the first one, done in https://github.com/pagopa/io-services-cms/pull/1579/changes/4b6a53d9d20a05b23e648c8f3058eab98d6c6e29

packages/external-clients/src/apim-client/index.ts
arcogabbo: thank you, I like the curried function suggestion, done in https://github.com/pagopa/io-services-cms/pull/1579/changes/4b6a53d9d20a05b23e648c8f3058eab98d6c6e29

rodoviti (approved):

PR #1585, opened 2026-06-16
Grant managed identity access and add identity-based settings

gunzip:
## Why

This is the safe first step of migrating Azure access from connection strings to managed identity. It only adds the permissions and non-breaking configuration required by the already-assigned managed identities.

## What changed

- Added least-privilege RBAC assignments for the CMS Function App production and staging slot managed identities on Storage Queue, Cosmos DB, and Event Hub resources using the DX `azure-role-assignments` module
- Added identity-based runtime/SDK settings under new prefixes for CMS Storage and Cosmos bindings/clients
- Added Event Hub namespace settings needed by the follow-up managed identity client migration
- Kept legacy connection-string settings for backward compatibility
- Preserved app-backend managed identity configuration as-is because it already uses managed identity for Blob Storage and Cosmos DB when no key is configured

## Migration map

- `app-backend` already uses `DefaultAzureCredential` for featured items Blob Storage and Cosmos DB AAD auth fallback; existing DX RBAC grants Storage Blob reader and Cosmos reader to production and staging slot identities.
- `io-services-cms-webapp` still uses connection strings for internal Storage Queue/Blob bindings, CMS Cosmos bindings, legacy Cosmos bindings, Event Hub producers, the Selfcare Event Hub trigger, and the legacy asset-storage SDK path. This PR only bootstraps identity settings and permissions.
- `INTERNAL_STORAGE_CONNECTION_STRING` is shared by application bindings and the Functions storage provider in `host.json`; it is intentionally kept as platform/runtime-required until a supported runtime replacement is explicitly verified.
- Staging slot async triggers remain disabled through sticky `AzureWebJobs.*.Disabled` settings. New identity endpoint settings are non-sticky because they describe the target resources and should swap consistently with application configuration.
- Secret ownership for external legacy Cosmos, Selfcare Event Hub, and external asset-storage connection strings is unclear, so this PR does not remove or mutate those secrets.

## Application order

Apply this PR first.

## Rationale

Deploying permissions before code changes reduces rollout risk: the application can continue using the current connection-string path while the required managed identity access is provisioned and propagated.

## Validation

- Ran `pre-commit run --all-files` manually. Terraform module locking, Terraform validate, and tflint passed after adding the missing Event Hub subscription data source.
- `terraform fmt` was run by pre-commit and updated Terraform formatting/lock data.
- VS Code diagnostics reported no errors on the Terraform files touched by this PR.

Note: the local Git hook did not inherit the terminal PATH for `pre-commit`, so the commit was created with `--no-verify` after running pre-commit manually. The local DX Technology Radar skill referenced by the workspace instructions was not available in this environment; this PR follows the existing repository DX module patterns.

### DIFF
--- a/infra/resources/_modules/cms_function_app/locals.tf
+++ b/infra/resources/_modules/cms_function_app/locals.tf
@@ -39,6 +39,8 @@ locals {
       COSMOSDB_NAME                           = "db-services-cms"
       COSMOSDB_APP_BE_NAME                    = "app-backend"
       COSMOSDB_URI                            = data.azurerm_cosmosdb_account.cosmos.endpoint
+      CMS_COSMOSDB__credential                = "managedidentity"
+      CMS_COSMOSDB__accountEndpoint           = data.azurerm_cosmosdb_account.cosmos.endpoint
       COSMOSDB_KEY                            = data.azurerm_cosmosdb_account.cosmos.primary_key
       COSMOSDB_CONTAINER_SERVICES_LIFECYCLE   = "services-lifecycle"
       COSMOSDB_CONTAINER_SERVICES_PUBLICATION = "services-publication"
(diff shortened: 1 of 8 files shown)


gunzip:
The latest iteration of this PR assigns full resource-level permissions for the Event Hub and Cosmos DB to the applications' managed identities. 

[A prior version](https://github.com/pagopa/io-services-cms/pull/1585/changes/9fb20ad2b20ab38c159e1938c1b6116972b0a2f9) implemented more granular permissions at the expense of a higher volume of role assignments.

This choice ultimately depends on the desired scope for each Function App. If a granular approach is preferred, we must ensure that new permissions are added whenever new connections are introduced.

giamma1295 (approved):
LGTM

PR #1586, opened 2026-06-16
Migrate Azure clients from connection strings to managed identity

gunzip:
## Why

This PR performs the actual application migration from connection strings to managed identity, with a temporary feature flag for safe rollback and local emulator support.

## What changed

- Added `USE_MANAGED_IDENTITY` feature flag to switch SDK clients and Node.js Azure Functions v4 code-defined bindings between managed identity and connection strings
- Switched CMS Cosmos clients and Event Hub producers to managed identity (`DefaultAzureCredential`) when the flag is enabled (default in deployed environments)
- Changed CMS Storage Queue/Blob and Cosmos binding registration so startup selects the new identity-based prefixes when the flag is enabled
- Added managed-identity credential verification to the `/api/v1/info` health check path used by the Function App
- Kept a connection-string fallback path gated behind the flag for quick rollback and local emulator development
- Updated local settings examples with `USE_MANAGED_IDENTITY=false`
- Added a Changeset because this PR changes runtime behavior for deployable applications

## Rollback procedure

- **SDK clients**: set `USE_MANAGED_IDENTITY=false` in app settings and restart. This reverts SDK clients to connection-string-based authentication without a code revert.
- **Bindings/triggers**: set `USE_MANAGED_IDENTITY=false` in app settings and restart/recycle the app or slot. On startup, the legacy binding configuration is registered again. The old connection-string settings from PR1 are still deployed, so bindings resume using the connection string without a code revert.

## Application order

Apply this PR after PR1.

Depends on #1585

## Validation strategy

- Production slot gets `USE_MANAGED_IDENTITY=true`; local examples set it to `false` for emulator/connection-string fallback.
- Staging slot async triggers remain disabled through sticky `AzureWebJobs.*.Disabled` settings. The `/api/v1/info` health check now performs pre-swap managed-identity probes against Storage Queue, CMS Cosmos, legacy Cosmos, and the services Event Hub namespace without enabling triggers.
- Because staging triggers are disabled, async trigger validation should be completed after swap through production observation and immediate rollback remains `USE_MANAGED_IDENTITY=false` plus app/slot restart.
- `INTERNAL_STORAGE_CONNECTION_STRING` remains in place because it is also used by the Functions storage provider/runtime in `host.json`.

## Validation

- `pnpm install`
- `pnpm --filter io-services-cms-webapp typecheck`
- `pnpm --filter io-services-cms-webapp test`
- `pnpm --filter io-services-cms-webapp lint` (passes with existing warnings only)
- `pnpm --filter io-services-app-backend typecheck`
- Git pre-commit hook: Terraform module locking skipped/no matching files where applicable, tflint passed, Terraform fmt passed, Terraform validate skipped/no matching files where applicable

## Rationale

This is the only behavior-changing step. By shipping it after the RBAC/bootstrap PR, the required identity permissions already exist. The feature flag provides a safety net for both SDK clients and Node.js Azure Functions v4 code-defined bindings. The startup credential verification catches RBAC misconfigurations before triggers start processing real messages.


apps/io-services-cms-webapp/src/webservice/controllers/info.ts
```diff
@@ -6,22 +10,80 @@ import {
...
-        ),
+        c.USE_MANAGED_IDENTITY
+          ? checkManagedIdentityHealth(c)
+          : healthcheck.checkAzureStorageHealth(
+              c.INTERNAL_STORAGE_CONNECTION_STRING,
+            ),
       (c) =>
-        healthcheck.checkAzureCosmosDbHealth(c.COSMOSDB_URI, c.COSMOSDB_KEY),
+        c.USE_MANAGED_IDENTITY
+          ? TE.right(true as const)
```
gunzip: I'd rather group these check for the connection string branch as in `checkManagedIdentityHealth` in some method ie. `checkConnStringHealth()` instead of returning `TE.right(true as const)`

apps/io-services-cms-webapp/src/config.ts
```diff
@@ -33,6 +34,119 @@ const InternalStorageAccount = t.type({
...
+  CMS_INTERNAL_STORAGE__blobServiceUri: NonEmptyString,
+  CMS_INTERNAL_STORAGE__queueServiceUri: NonEmptyString,
+  CMS_LEGACY_COSMOSDB__accountEndpoint: NonEmptyString,
+  SERVICES_EVENT_HUB_FULLY_QUALIFIED_NAMESPACE: NonEmptyString,
+};
+
+const ManagedIdentityOptionalSettings = t.partial(ManagedIdentitySettings);
+const ManagedIdentityRequiredSettings = t.type(ManagedIdentitySettings);
+
+const FallbackSettings = {
```
gunzip, line 53: please add a comment about the reson we keep these fallback settings

apps/io-services-cms-webapp/src/config.ts
```diff
@@ -33,6 +34,119 @@ const InternalStorageAccount = t.type({
...
+  (
+    input,
+  ): input is
+    | RuntimeModeDisabledConfiguration
+    | RuntimeModeEnabledConfiguration =>
+    typeof input === "object" && input !== null,
+  (input, context) => {
+    const useManagedIdentityOrErrors = ManagedIdentityFlag.validate(
+      typeof input === "object" && input !== null
+        ? (input as Record<string, unknown>).USE_MANAGED_IDENTITY
```
gunzip: if we really need a cast here, better use 'unknown'

apps/io-services-cms-webapp/src/main.ts
```diff
@@ -267,39 +271,88 @@ const legacyServiceModel = new ServiceModel(legacyServicesContainer);
 
 const blobService = createBlobService(config.ASSET_STORAGE_CONNECTIONSTRING);
 
+const defaultAzureCredential = new DefaultAzureCredential();
+const runtimeModeConfiguration:
+  | RuntimeModeDisabledConfiguration
+  | RuntimeModeEnabledConfiguration = config;
+const useManagedIdentity = runtimeModeConfiguration.USE_MANAGED_IDENTITY;
+
+const requireFallbackSetting = (
```
gunzip: do we really need this method `requireFallbackSetting` just to satisfy typings or can we rely on better schemas and inference at compile time?

apps/app-backend/local.settings.json.example
```diff
@@ -2,10 +2,11 @@
   "IsEncrypted": false,
   "Values": {
     "FUNCTIONS_WORKER_RUNTIME": "node",
+    "USE_MANAGED_IDENTITY": "false",
```
giamma1295: Why is necessary this env in `app-backend`, given that no changes have been made to the application?

This application is already subject to RBAC authentication (Cosmos and storageAccount), the only service where we use keys is Azure AI Search, but it hasn't been modified.

infra/resources/_modules/function_app/locals.tf
```diff
@@ -3,7 +3,8 @@ locals {
     use_case      = "high_load"
     cosmosdb_name = "app-backend"
     app_settings = {
-      NODE_ENV = "production"
+      NODE_ENV             = "production"
+      USE_MANAGED_IDENTITY = "true"
```
giamma1295: Same as `apps/app-backend/local.settings.json.example`

.changeset/managed-identity-migration.md
```diff
@@ -0,0 +1,6 @@
+---
+"io-services-cms-webapp": patch
+"io-services-app-backend": patch
+---
+
+Migrate Azure client and binding configuration toward managed identity with a temporary rollback flag.
```
giamma1295: nit: Isn't the mechanism for switching between ManagedIdentity and ConnectionString permanent?

When testing locally using emulators, will it still be necessary to connect using a connection string in the future!

giamma1295:
`io-services-cms-webapp/host.json` contains:

```json
"storageProvider": {
      "connectionStringName": "INTERNAL_STORAGE_CONNECTION_STRING",
      "partitionCount": 4
    }
```
Is this still needed?

apps/io-services-cms-webapp/src/webservice/controllers/info.ts
```diff
@@ -6,22 +10,100 @@ import {
...
+        config.COSMOSDB_KEY,
+      ),
+    ),
+  );
+
+const checkManagedIdentityHealth = (
+  config: ManagedIdentityHealthConfiguration,
+): healthcheck.HealthCheck<"AzureManagedIdentity"> =>
+  TE.tryCatch(async () => {
+    const credential = new DefaultAzureCredential();
```
giamma1295: DefaultAzureCredential is instantiated every time the health check task runs(every time /info API is invoked), consider wire as dependency in `main.ts`

gunzip:
@giamma1295 regarding `storageProvider`, I think this setting is currently being ignored. It needs to be placed under the `extensions.durableTask.storageProvider` path to be valid.

apps/io-services-cms-webapp/src/webservice/controllers/info.ts
gunzip: good catch!

giamma1295 (approved):
LGTM

PR #1590, opened 2026-06-24
[IOPID-3989] Make `selcGroups` required in `BackOfficeUserPermissions`

adelloste:
This PR makes the `selcGroups` in `BackOfficeUserPermissions` a required array (defaulting to `[]` when no groups are present), instead of being possibly `undefined`


rodoviti (approved):
lgtm

PR #1618, opened 2026-07-20
Add Zod Schemas for ServicesLifecycle, ServicePublication e ServiceTopic

giamma1295:
Closes IOCOM-3270

### DIFF
--- a/apps/services-app/src/domain/entities/service-publication.ts
+++ b/apps/services-app/src/domain/entities/service-publication.ts
@@ -2,16 +2,10 @@ import z from "zod";
 
 import { serviceSchema } from "./service.js";
 
-const publicationFsmSchema = z.discriminatedUnion("state", [
-  z.object({
-    lastTransition: z.string().optional(),
-    state: z.literal("published"),
-  }),
-  z.object({
-    lastTransition: z.string().optional(),
-    state: z.literal("unpublished"),
-  }),
-]);
+const publicationFsmSchema = z.object({
+  lastTransition: z.string().optional(),
+  state: z.union([z.literal("published"), z.literal("unpublished")]),
+});
 
 export const servicePublicationSchema = serviceSchema.extend({
   fsm: publicationFsmSchema,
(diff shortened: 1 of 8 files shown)


apps/services-app/src/domain/entities/service-publication.ts
```diff
@@ -0,0 +1,20 @@
...
+const publicationFsmSchema = z.discriminatedUnion("state", [
+  z.object({
+    lastTransition: z.string().optional(),
+    state: z.literal("published"),
+  }),
+  z.object({
+    lastTransition: z.string().optional(),
+    state: z.literal("unpublished"),
+  }),
+]);
```
Garma00: Since there is no difference between the two types, can we just define `publicationFsmSchema` as a `z.object`? Something like this:
```suggestion
const publicationFsmSchema =   z.object({
    lastTransition: z.string().optional(),
    state: z.union([z.literal("published"), z.literal("unpublished")]),
  }),
```

apps/services-app/src/domain/entities/service-topic.ts
```diff
@@ -0,0 +1,8 @@
+import z from "zod";
+
+export const serviceTopicSchema = z.object({
+  id: z.number().int(),
```
Garma00:
```suggestion
  id: z.int(),
```

Garma00 (approved):
LGTM I just left those 2 possible improvements, feel free to ignore them.

apps/services-app/src/domain/entities/service-topic.ts
giamma1295: done in 6b66d70e87a35e8d57b2cd24d2b5b11279b30a61 !

apps/services-app/src/domain/entities/service-publication.ts
giamma1295: You are absolutely right! done in baeb8abc117551ffa19065e19de50ee9c9fe277a !

PR #1619, opened 2026-07-20
Add GetServiceInternal UseCase and Dependant ports

giamma1295:
Closes IOCOM-3271

### DIFF
--- a/apps/services-app/src/application/ports/service-lifecycle-repository.ts
+++ b/apps/services-app/src/application/ports/service-lifecycle-repository.ts
@@ -2,7 +2,17 @@ import type { ServiceLifecycle } from "@/domain/entities/service-lifecycle.js";
 import type { GenericError, NotFoundError } from "@pagopa/hexagonal-core";
 import type { Result } from "neverthrow";
 
+/**
+ * Provides access to services in their lifecycle state.
+ */
 export interface ServiceLifecycleRepository {
+  /**
+   * Retrieves a lifecycle service by its identifier.
+   *
+   * @param serviceId - The unique identifier of the service.
+   * @returns The lifecycle service, a `NotFoundError` when it does not exist,
+   * or a `GenericError` when the repository cannot complete the operation.
+   */
   get(
     serviceId: string,
   ): Promise<Result<ServiceLifecycle, GenericError | NotFoundError>>;
--- a/apps/services-app/src/application/ports/service-lifecycle-repository.ts
+++ b/apps/services-app/src/application/ports/service-lifecycle-repository.ts
@@ -1,0 +1,9 @@
+import type { ServiceLifecycle } from "@/domain/entities/service-lifecycle.js";
+import type { GenericError, NotFoundError } from "@pagopa/hexagonal-core";
+import type { Result } from "neverthrow";
+
+export interface ServiceLifecycleRepository {
+  get(
+    serviceId: string,
+  ): Promise<Result<ServiceLifecycle, GenericError | NotFoundError>>;
+}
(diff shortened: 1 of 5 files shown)


apps/services-app/src/application/ports/service-lifecycle-repository.ts
```diff
@@ -0,0 +1,9 @@
+import type { ServiceLifecycle } from "@/domain/entities/service-lifecycle.js";
+import type { GenericError, NotFoundError } from "@pagopa/hexagonal-core";
+import type { Result } from "neverthrow";
+
+export interface ServiceLifecycleRepository {
+  get(
+    serviceId: string,
+  ): Promise<Result<ServiceLifecycle, GenericError | NotFoundError>>;
```
Garma00, line 18: It could be useful to document the repository directly, similar to what we did in [io-messages](https://github.com/pagopa/io-messages/blob/main/apps/messages-app/src/application/ports/message-metadata.ts#L29-L29). That way, the documentation is shared and consistent for any adapter that implements it. I don't have a strong preference either way, so feel free to skip it if you prefer.

apps/services-app/src/application/ports/service-publication-repository.ts
```diff
@@ -0,0 +1,9 @@
+import type { ServicePublication } from "@/domain/entities/service-publication.js";
+import type { GenericError, NotFoundError } from "@pagopa/hexagonal-core";
+import type { Result } from "neverthrow";
+
+export interface ServicePublicationRepository {
```
Garma00, line 8: Ditto

apps/services-app/src/application/ports/topic-repository.ts
```diff
@@ -0,0 +1,9 @@
+import type { ServiceTopic } from "@/domain/entities/service-topic.js";
+import type { GenericError, NotFoundError } from "@pagopa/hexagonal-core";
+import type { Result } from "neverthrow";
+
+export interface TopicRepository {
```
Garma00, line 8: Ditto

Garma00 (approved):
LGTM, I just left some non-blocking suggestions.

apps/services-app/src/application/ports/service-lifecycle-repository.ts
daniacca, line 18: Since this type is quite recurrent in our Repository port:

`Result<ServiceLifecycle, GenericError | NotFoundError>`

We could define an alias and reuse the alias everywhere, like:

```
import type { GenericError, NotFoundError } from "@pagopa/hexagonal-core";
import type { Result } from "neverthrow";

export type RepositoryResult<T> = Result<T, GenericError | NotFoundError>;
```

daniacca (approved):

apps/services-app/src/application/ports/service-lifecycle-repository.ts
giamma1295: ts-doc added in c888822eed5812a92f3e0d339b0dce51c7133894

apps/services-app/src/application/ports/service-publication-repository.ts
giamma1295: ts-doc added in https://github.com/pagopa/io-services-cms/commit/c888822eed5812a92f3e0d339b0dce51c7133894

apps/services-app/src/application/ports/topic-repository.ts
giamma1295: ts-doc added in https://github.com/pagopa/io-services-cms/commit/c888822eed5812a92f3e0d339b0dce51c7133894

apps/services-app/src/application/ports/service-lifecycle-repository.ts
giamma1295: Good suggestion, but at that moment we are not 100% sure if future methods in these repos will maintains the same signature for errors, so we can make it generic in future development!

daniacca (approved):

